PR #5, opened 2022-09-28
Fix changing boot devices order for HyperCore v9.2

justinc1:
The VM uuid is already in PATCH URL path, it is not needed in PATCH data/payload.
v9.1 did not complain, v9.2 does complain.

Fix by @ddemlow


justinc1:
The integration tests for HC3 v9.1 are still running, so fall all looks ok.

dradX (approved):
Seems OK to me.

domendobnikar (approved):

justinc1:
`ansible-test integration` - passed

PR #13, opened 2022-09-29
Sanity fix after commiting issue 7

domendobnikar:
There was some sanity issue left out and this is a fix for it.


justinc1:
Thank  you

PR #14, opened 2022-09-29
allow VM with 0 vCPUs

domendobnikar:


plugins/module_utils/vm.py
```diff
@@ -883,7 +883,9 @@ def set_vm_params(module, rest_client, vm):
             endpoint = "{0}/{1}".format("/rest/v1/VirDomain", vm.uuid)
             task_tag = rest_client.update_record(endpoint, payload, module.check_mode)
             TaskTag.wait_task(rest_client, task_tag)
-            if ManageVMParams._needs_reboot(module, changed_parameters) and vm.power_state not in ["stop", "stopped", "shutdown"]:
```
justinc1, line 886: sanity test - line length, need to check

justinc1: Is ok :)

PR #16, opened 2022-09-29
Update Readme.md

justinc1:
It is directly included at
https://galaxy.ansible.com/scale_computing/hypercore, we need to be compatible.


dradX (approved):
Seems good to me.

PR #17, opened 2022-09-29
Gh actions

justinc1:
PR add ansible-test sanity, units and integration, and runs ansible-lint.

@ddemlow Please check if implemented CI is compatible with Scale Computing rules.


justinc1:
The CI tests / integration-test  failed because SMB server is not jet correctly setup - we need to check and fix username/password.
The github action file is OK, so this should be merged. @dradX

dradX (approved):

PR #20, opened 2022-09-30
Replace '=' with '__' (double underscore) in tags recognized by inventory plugin

justinc1:
The HyperCore web UI does not allow tag with '=', so we must not use it.


domendobnikar (commented):
I think this needs to be changed as well.
```
  for tag in tags:
                    if tag.startswith("ansible_host="):
                        ansible_host = tag[len("ansible_host=") :]
```
And inside integration test as well.

justinc1:
Thank you @domendobnikar for fixing the ansible_host - I just missed it.

PR #21, opened 2022-09-30
Link for docs

justinc1:
This is what is shown on https://galaxy.ansible.com/scale_computing/hypercore, so we need clickable links.


dradX (approved):
Thanks for these docs changes @justinc1 and setting up github pages @ddemlow

PR #22, opened 2022-09-30
Fixing tiering priority:

domendobnikar:
- Added a mapper to back end values as provided by Dave
- VM should not reboot anymore when disk tiering priority is being changed


justinc1:
Looking at CI - https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/3160152672/jobs/5145386784, 
```
PLAY [Cleanup test environment] ************************************************

TASK [List all VMs] ************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 12
fatal: [localhost]: FAILED! => {"changed": false, "module_stderr": "Traceback (most recent call last):\n  File \"/root/.ansible/tmp/ansible-tmp-1664560391.5457187-994-60392471886543/AnsiballZ_vm_info.py\", line 121, in ...
```
Looks like the vm_info module is broken?

plugins/module_utils/disk.py
```diff
@@ -113,7 +146,11 @@ def from_ansible(cls, ansible_dict):
             cache_mode=ansible_dict.get("cache_mode", None),
             name=ansible_dict.get("iso_name", None),
             disable_snapshotting=ansible_dict.get("disable_snapshotting", None),
-            tiering_priority_factor=ansible_dict.get("tiering_priority_factor", None),
+            tiering_priority_factor=TIERING_PRIORITY_MAPPING_FROM_ANSIBLE[
+                ansible_dict["tiering_priority_factor"]
+            ]
+            if ansible_dict.get("tiering_priority_factor", None) is not None
+            else None,
```
justinc1, line 153: I think, as a rule, we are storing into our ansible classes the ansible-friendly values. So the tiering_priority_factor should contain numbers 0 to 11, and to HC3 API we send "0, 1, 2,... 1024, 10240".

tests/integration/targets/vm_git_issues/tasks/issue_15_tiering_priority.yml
```diff
@@ -0,0 +1,155 @@
...
+
+- name: Retrieve XLAB-test-tiering-priority
+  scale_computing.hypercore.vm_info:
+    vm_name: XLAB-test-tiering-priority
+  register: source_info
+- ansible.builtin.assert:
+    that:
+      - source_info.records | length == 1
+      - source_info.records.0.power_state == "stopped"
+      - source_info.records.0.disks.0.tiering_priority_factor == 0
```
justinc1, line 42: Here I think - maybe tiering_priority_factor should be 4 by default?

tests/integration/targets/vm_git_issues/tasks/issue_15_tiering_priority.yml
```diff
@@ -0,0 +1,155 @@
...
+
+- name: Get actual tiering_priority values from hypercore without mapping
+  scale_computing.hypercore.api:
+    action: get
+    endpoint: /rest/v1/VirDomain/{{vm_created.record.createdUUID}}
+  register: result
+- ansible.builtin.assert:
+    that:
+      - result.record.0.name == "XLAB-test-tiering-priority"
+      - result.record.0.blockDevs.0.tieringPriorityFactor == 32
```
justinc1, line 113: Nice you check also this raw API value :)

justinc1 (changes requested):
The default value for `tiering_priority_factor` should be same as set by HyperCore. I created a new VM via web UI, and the created disk was:
```
                "tieringPriorityFactor": 8,
```

So 8 on HyperCores side == 4 on Ansible side.

tests/integration/targets/vm_git_issues/tasks/issue_15_tiering_priority.yml
domendobnikar: Yes, I think so, I'm not exactly sure, if a newly created VM has tiering_priority_factor set to 4 as default or 0.

justinc1: This is fixed in test now. Problem is HC3 returs 0 for first 5 secods, after 30 sec or more it does return expected 8

justinc1 (approved):

justinc1:
gitlab CI - integration test passed there - https://gitlab.xlab.si/scale-ansible-collection/HyperCoreAnsibleCollection/-/pipelines/37431

PR #23, opened 2022-09-30
Add example playbooks for vm_info and vm_replication_info modules

justinc1:
Nothing special in those two playbooks, they are just ready-to-run example playbooks.

@ddemlow Are such examples something that you would find useful when you craft your own playbooks? The `examples/` were added with intention that endusers could run then immediately (ok, after installing ansible and `export SC_HOST=...`.


justinc1:
I just noticed vm_replication_info cannot report status for VM named `test-tiering`. The VM replication and snapshot schedule are both uncofigured.

PR #32, opened 2022-10-04
3 enlarging virtual disk is triggering vm restart cycle

domendobnikar:


justinc1:
Code is OK. I extended integration test a bit - just to check a few things/variables that can never have a value different from what we expect it to be.

PR #43, opened 2022-10-19
Bugfix: we do not crash if machineType is empty

juremedvesek:
For some instances API returns VirDomain with empty machineType. Ansible should work with instances like that - at least read part should not have any problems.


justinc1 (commented):
Looks as correct thing.

PR #49, opened 2022-11-10
api module: Add support for PUT methods

TomboScaleComputing:


plugins/module_utils/rest_client.py
```diff
@@ -84,20 +84,19 @@ def put_record(
...
         binary_data=None,
         headers=None,
     ):
         # Method put doesn't support check mode # IT ACTUALLY DOES
         if check_mode:
             return None
-        # Only /rest/v1/ISO/[uuid}/data is using put, which doesn't return anything.
-        # self.client.put on this endpoint returns None.
         try:
             response = self.client.put(
```
domendobnikar, line 96: `response `needs to be json when it's being returned, since `put_record()` is then returning it directly to` main()` which expects it to be json.
So `self.client.put().json` or `return response.json`.

domendobnikar (approved):

plugins/module_utils/rest_client.py
TomboScaleComputing: @domendobnikar - currently the only other place our API uses a PUT request is for the ISO upload. That endpoint (unlike the new one for VirtualDisk) does not return json so I think this needs to be checked before returning from this function?

PR #51, opened 2022-11-30
41 cloud init disk logic fix

domendobnikar:
Allows for cloud init to create IDE_CDROM without deleting it as long as cloud-init ISO is still attached.
Removes cloud init CDROM when ISO is detached.

### DIFF
--- a/tests/integration/targets/vm/tasks/01_main.yml
+++ b/tests/integration/targets/vm/tasks/01_main.yml
@@ -32,7 +32,7 @@
         state: present
       register: result
 
-    - name: Create and start the VM with disks, nics and boot devices set. Attach ISO onto the VM. Add cloud init data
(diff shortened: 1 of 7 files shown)


tests/integration/targets/vm/tasks/01_main.yml
```diff
@@ -62,16 +62,6 @@
             disk_slot: 0
           - type: nic
             nic_vlan: 1
-        cloud_init:
```
domendobnikar, line 65: This was moved to another integration test, since cloud_init wasn't working correctly at the time, this test was ignoring the cloud_init value.
To keep the integrity of the test the same as it was before the change, I removed it from here and made a separate test for it.

justinc1 (approved):
Nice work.

I only extended integration tests a bit - check ISO name is in correct slot, and case when cloud-init data is present, but ISO was already manually ejected from CD-ROM and CD-ROM removed.

PR #53, opened 2022-12-07
SMB cleanup script

domendobnikar:


tests/integration/cleanup/smb_cleanup.sh
```diff
@@ -0,0 +1,53 @@
...
+ }
+
+# Main function
+main () {
+    # $1 server address
+    # $2 share on server
+    # $3 username
+    # $4 password
+    # username is provided as domain;username
+    IFS=';' read -ra username <<< $3
```
justinc1, line 42: Nice. I didn't know about '<<<' operator until now.

PR #54, opened 2022-12-08
Attach_guest_tools: operating_system field added to vm module and vm_params

domendobnikar:
There was a problem where `attach_guest_tools` wasn't working, because operating system wasn't being specified during VM creation.
Adding OS field now lets user define which system is being used, user can also change it with `vm_params` as per request from @ddemlow .
`attach_guest_tools` only works with Windows systems.


tests/integration/targets/vm/tasks/02_shutdown_cases.yml
```diff
@@ -36,7 +36,7 @@
       - mytag2
     memory: "{{ '512 MB' | human_to_bytes }}"
     vcpu: 2
-    attach_guest_tools_iso: true
+    attach_guest_tools_iso: false
```
domendobnikar, line 39: `attach_guest_tools_iso` wasn't working correctly at the time when this test was written.
I changed the value to `false` everywhere where this was the case to keep the test integrity.
Made a separate test in the vm_git folder to test `attach_guest_tools_iso`.

PR #58, opened 2022-12-12
Update main.yml - set mem on name: Create XLAB-vm_clone-CI_test-running

ddemlow:
made mem == 511705088 change to name: Create XLAB-vm_clone-CI_test-running  (which was the one not starting on 9.2.8 hypercore)

### DIFF
--- a/tests/integration/targets/vm_clone/tasks/main.yml
+++ b/tests/integration/targets/vm_clone/tasks/main.yml
@@ -91,7 +91,7 @@
           dom:
             name: XLAB-vm_clone-CI_test-running
             tags: Xlab,CI,test,vm_clone,running
-            mem: 512100100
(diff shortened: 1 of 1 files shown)


justinc1:
One question @ddemlow : "which was the one not starting on 9.2.8 hypercore" - does this mean that on 9.2 VM does not start if mem is not rounded to 1 MB?

ddemlow:
it was XLAB-vm_clone-CI_test-running that wouldn't start and fail the test  (my first commit I changed memory on the wrong VM) ... yes there is a bug specific to 9.2.8 that the VM won't start if ram bytes are not rounded to 1MiB size.   9.1 and 9.2.7 is fine - and that bug is expected to be fixed in hypercore however I'm not sure yet what release

PR #62, opened 2022-12-21
Implementing missing Timeout logic

domendobnikar:


plugins/module_utils/client.py
```diff
@@ -63,10 +64,20 @@ def __init__(
         self.host = host
         self.username = username
         self.password = password
+        self.timeout = timeout
 
         self._auth_header = None
         self._client = Request()
 
+    @classmethod
+    def get_client(cls, cluster_instance: dict):
```
domendobnikar, line 73: Added this `classmethod` in order to avoid writing the same code in every module.
Plus it generates an instance of class Client.

plugins/modules/api.py
```diff
@@ -282,11 +282,7 @@ def main():
     )
 
     try:
-        client = Client(
-            host=module.params["cluster_instance"]["host"],
-            username=module.params["cluster_instance"]["username"],
-            password=module.params["cluster_instance"]["password"],
-        )
+        client = Client.get_client(module.params["cluster_instance"])
```
domendobnikar, line 285: Made this into a `classmethod` in every module ... seems more appropriate since it's creating an instance of a class anyways.

plugins/module_utils/client.py
justinc1: Nice, I aggree

justinc1:
Thank you, nice.

PR #64, opened 2022-12-22
Updated snapshot schedule integration tests

domendobnikar:


tests/integration/targets/snapshot_schedule/tasks/main.yml
```diff
@@ -5,21 +5,6 @@
     SC_PASSWORD: "{{ sc_password }}"
 
   block:
-    - name: Retrieve all snapshot schedules
-      scale_computing.hypercore.snapshot_schedule_info:
-      register: initial
-    - ansible.builtin.assert:
-        that:
-          - initial.records != []
```
domendobnikar, line 13: I don't think we need to have this test, it just opens up potential issues.
Check if any schedules exists, if not fail the test ... seems completely pointless to have this.
It's always going to fail on a new cluster and it doesn't check anything valuable.

tests/integration/targets/snapshot_schedule/tasks/main.yml
```diff
@@ -5,21 +5,6 @@
...
-
-    - name: Retrieve specific snapshot schedule object by name and verify that query for that name was successful.
-      scale_computing.hypercore.snapshot_schedule_info:
-        name: "{{ initial.records.0.name }}"
-      register: result
-    - ansible.builtin.assert:
-        that:
-          - result.records.0.name == initial.records.0.name
```
domendobnikar, line 21: This is just the continuation of the previous test, again completely pointless to have this.
We are already checking if info module works ... The same goes for creation/deletion/update of the schedule.
This is again always going to fail on a new cluster, seems to add no value to the actual integration test.

domendobnikar: Hmm, is there a difference between replication created schedule vs user created schedule?
If there is than running a mock replication at the start of this test would 100% make sense, make it a part of this test.
I would however prefer to avoid having any kind of dependency between different integration tests.

ddemlow: avoiding dependencies makes sense ... 

but  no difference ... usually default rules are to replicate every 15 min... but our replication code will create that schedule if it doesn't already exist  (and I guess this one is hidden from U schedules list) ... 

UUID            : c4f652b1-9cf7-4171-817f-d2b24994ae5c
Name            : Replication Default
Recurrence Rules: Name              UUID                                  Local DTSTART        Replication  Local Retain  Remote Retain  RRULE
                  ================  ====================================  ===================  ===========  ============  =============  =========================
                  Every 15 Minutes  c3e2cab7-3bf8-46b1-abac-2e116411f97b  2016-01-01 00:00:00  yes          -             -              FREQ=MINUTELY;INTERVAL=15
============================================================================== ROW 3 ==============================================================================

justinc1: I agree there is not much need to check what (if any) snapshot schedule is already present. We can remove those two checks for initial state , and test only what snapshot_schedule module does.

PR #70, opened 2023-01-05
Timeout integration config

domendobnikar:
Adding logic that allows for timeout to be read from integration_config and github secret.
This is an update that effects integration tests only.

### DIFF
--- a/plugins/module_utils/arguments.py
+++ b/plugins/module_utils/arguments.py
@@ -38,7 +38,7 @@ SHARED_SPECS = dict(
             timeout=dict(
                 type="float",
                 required=False,
-                no_log=False,
+                no_log=True,
                 fallback=(env_fallback, ["SC_TIMEOUT"]),
             ),
         ),
--- a/plugins/module_utils/arguments.py
+++ b/plugins/module_utils/arguments.py
@@ -36,9 +36,9 @@ SHARED_SPECS = dict(
                 fallback=(env_fallback, ["SC_PASSWORD"]),
             ),
             timeout=dict(
-                type="float",
+                type="str",
                 required=False,
-                no_log=True,
+                no_log=False,
                 fallback=(env_fallback, ["SC_TIMEOUT"]),
             ),
         ),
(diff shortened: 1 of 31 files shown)


tests/integration/targets/api/tasks/main.yml
```diff
@@ -3,6 +3,7 @@
     SC_HOST: "{{ sc_host }}"
     SC_USERNAME: "{{ sc_username }}"
     SC_PASSWORD: "{{ sc_password }}"
+    SC_TIMEOUT: "{{ sc_timeout|default(None) or 10 }}"
```
domendobnikar, line 6: Only default wasn't working here.
Timeout can be empty string instead of None and is treated as a string.

plugins/module_utils/arguments.py
```diff
@@ -38,7 +38,6 @@
             timeout=dict(
                 type="float",
                 required=False,
-                no_log=True,
```
domendobnikar, line 41: For some reason this was effecting vm_info module?
I don't know how it's possible, but certain values were returned as *****.

justinc1: If timeout had value "bla", than every "bla" in ansible output would be replaced by ****. Maybe timeout had value "0" or "None", those are common, so a lot of unrelated text would be replaced.

justinc1: Nice you removed that line.

tests/integration/targets/api/tasks/main.yml
justinc1: I think "{{ sc_timeout|default(10) }}" (or "{{ sc_timeout|default('10') }}" ) would also work.

domendobnikar: I tried that first, but if `sc_timeout` can be `' '` instead of `None `- which is treated as a `string `and not `float`.
I was getting `"cannot convert string to float exception"`.

justinc1: maybe problem is that we have no `secrets.SC_TIMEOUT` in github.

justinc1 (approved):

PR #74, opened 2023-01-18
Us01 user info module

PolonaM:
Added user_info module

### DIFF
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2023, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from ..module_utils.utils import PayloadMapper
+
+
+class User(PayloadMapper):
+    def __init__(self, uuid, username, fullname, role_uuids, session_limit):
+        self.uuid = uuid
+        self.username = username
+        self.fullname = fullname
+        self.role_uuids = role_uuids
+        self.session_limit = session_limit
+
+    @classmethod
+    def from_ansible(cls):
+        pass
+
+    @classmethod
+    def from_hypercore(cls, user_dict):
+        if not user_dict:  # In case for get_record, return None if no result is found
+            return None
+        return cls(
+            uuid=user_dict["uuid"],
+            username=user_dict["username"],
+            fullname=user_dict["fullName"],
+            role_uuids=user_dict["roleUUIDs"],
+            session_limit=user_dict["sessionLimit"],
+        )
+
+    def to_hypercore(self):
+        pass
+
+    def to_ansible(self):
+        return dict(
+            uuid=self.uuid,
+            username=self.username,
+            fullname=self.fullname,
+            role_uuids=self.role_uuids,
+            session_limit=self.session_limit,
+        )
+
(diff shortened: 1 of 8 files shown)


plugins/modules/user_info.py
```diff
@@ -0,0 +1,81 @@
...
+records:
+  description:
+    - A list of user records.
+  returned: success
+  type: list
+  sample:
+    - fullname: xlab
+      role_uuids:
+        - 38b346c6-a626-444b-b6ab-92ecd671afc0
+        - 7224a2bd-5a08-4b99-a0de-9977089c66a4
```
justinc1, line 42: Role uuid does not tell much. I guess later we will need to add also a role name. But dedicated role_info module is out of scope in this phase, so I want only to have format that will allow future modification in backward compatible way.

If today we output:
```
  sample:
    - fullname: xlab
      roles:
        - uuid: 38b346c6-a626-444b-b6ab-92ecd671afc0
        - uuid: 7224a2bd-5a08-4b99-a0de-9977089c66a4
```

And in future we can extend with (role) name:
```
  sample:
    - fullname: xlab
      roles:
        - uuid: 38b346c6-a626-444b-b6ab-92ecd671afc0
          name: admin
        - uuid: 7224a2bd-5a08-4b99-a0de-9977089c66a4
          name: backup
```

plugins/module_utils/user.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
```
justinc1, line 1: File included by mistake? Then move it to next PR.

plugins/modules/user_info.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def main():
+    module = AnsibleModule(
+        supports_check_mode=True,
+        argument_spec=dict(arguments.get_spec("cluster_instance")),
+    )
+
+    try:
+        client = Client.get_client(module.params["cluster_instance"])
```
justinc1, line 71: We have also a CachedRestClient - it is intended for `_info` modules, it does read caching. It will not improve performance here (a single HTTP GET), but we should use it where it makes sense.

justinc1 (changes requested):

plugins/modules/user_info.py
justinc1: Update @PolonaM: please include role name in user_info output. Also when creating a new user, roles will need to be selected, and role will be identified by name.

plugins/modules/user_info.py
PolonaM: @justinc1 Thanks for the tip! But as you said, in the user_info module is not needed, so I used RestClient.

tests/integration/targets/user_info/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+    SC_TIMEOUT: "{{ sc_timeout }}"
+
+  block:
+    - name: Retrieve info about the users
+      scale_computing.hypercore.user_info:
+      register: users
+    - ansible.builtin.assert:
+        that:
+          - users.records != []
+          - "{{ users.records[0].keys() | sort == ['fullname', 'roles', 'session_limit', 'username', 'uuid'] }}"
```
justinc1, line 15:
```
(users.records[0].keys() | sort) == ['full_name', 'roles', 'session_limit', 'username', 'uuid']
```

justinc1 (changes requested):
minor changes

tests/integration/targets/user_info/tasks/main.yml
PolonaM: Tanks for the comment, integration test fixed

justinc1 (approved):
Thank you :)

PR #79, opened 2023-01-26
US01 User module added 

PolonaM:


plugins/modules/user.py
```diff
@@ -0,0 +1,261 @@
...
+from ..module_utils import arguments, errors
+from ..module_utils.client import Client
+from ..module_utils.rest_client import RestClient
+from ..module_utils.rest_client import CachedRestClient
+from ..module_utils.user import User
+from ..module_utils.role import Role
+
+
+def get_role_uuids(module, rest_client: RestClient):
+    # CachedRestClient is beneficial here since we use for loop and make get requests to the same endpoint many times
```
justinc1, line 122: Oh no :) The CachedRestClient is intended to cache GET responses. It does not know when to flush cached data - so you can easily get outdated data from it. Until this is cache flush is implemented, it cannot be used in `wait until object.get_state()=="something"` . Please do not use it action modules.

If your code does "read user, modify user, read user", and REST calls are "GET users/uuid1, PUT users/uuid1, GET users/uuid1", the second GET will return data from first GET. Unless you manually flush cache by creating a new instance of CachedRestClient?

Or your code does not do second GET, as you can just assume new server state is same to PUT payload.

Eventually we can add better cache options:
- flush after each PUT/POST/DELETE
- flush manually
- flush after entry expires, based on time.

justinc1: Or maybe we do keep it here? It is not problem you are using CachedRestClient, problem is Jure and me didn't make it good enough. Integration tests do ensure this module DOES work. And if we can spare unneeded GET request, without making code hard to read - this is a good thing.

Just keep using it. Module does work. CachedRestClient can be made better too.

Just FYI. CC @domendobnikar  @anazobec

tests/integration/targets/user/tasks/main.yml
```diff
@@ -0,0 +1,115 @@
...
+          - Admin
+        session_limit: 2
+      register: user
+    - ansible.builtin.assert:
+        that:
+          - user is changed
+          - user.record.keys() | sort == ['full_name', 'roles', 'session_limit', 'username', 'uuid']
+          - user.record.username == "my_user"
+          - user.record.full_name == "my_full_name"
+          - user.record.roles[0]["name"] == "Admin"
```
justinc1, line 33: check len(user.record.roles) == 1

tests/integration/targets/user/tasks/main.yml
```diff
@@ -0,0 +1,115 @@
...
+          - user.record.session_limit == 2
```
justinc1, line 34: You should use also `user_info` module to check what is actual stat on server. Checking content of `user` variable only checks what user module thinks was setup on the HC3.

I need to look into python code to see if you read new state from server or not. And if remove the line that does it - the test would not detect it.

plugins/modules/user.py
```diff
@@ -0,0 +1,261 @@
...
+        user,
+        dict(before=user, after=user),
+    )
+
+
+def delete_user(rest_client: RestClient, user):
+    if user:
+        user.delete(rest_client)
+        return True, dict(), dict(before=user.to_ansible(rest_client), after={})
+    return False, dict(), dict(before={}, after={})
```
justinc1, line 203: I almost comment why you return `dict(before={}, ` (line 203 - last one) if you deleted a user?

A bit easier to read if special case of user not present is hadled first, the the normal code excution does not need to be in the `if`.

```
    if not user:
        return False, dict(), dict(before={}, after={})
    user.delete(rest_client)
    return True, dict(), dict(before=user.to_ansible(rest_client), after={})
```

justinc1 (changes requested):
minor changes. thank you. :)

plugins/modules/user.py
```diff
@@ -0,0 +1,261 @@
...
+  username:
+    description:
+      - The user name.
+      - Serves as unique identifier.
+      - If C(state) is I(present) and C(username) doesn't exist, a new user is created.
+      - If C(state) is I(present) and C(username) already exists, existing user is updated.
+      - If C(state) is I(absent) and C(username) is found, existing user is deleted.
+    type: str
+    required: True
+  new_username:
```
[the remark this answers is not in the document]
PolonaM: @justinc1 Thanks for the comment, I really did forget to recheck documentation before PR. I updated the new username option and also did some other corrections to the documentation.

tests/integration/targets/user/tasks/main.yml
PolonaM: Added check

plugins/modules/user.py
PolonaM: Changed the code order as suggested

tests/integration/targets/user/tasks/main.yml
PolonaM: Thanks for the comment, I agree with you, I added checks with user_info module

plugins/modules/user.py
PolonaM: @justinc1 There is no case in the module where there is "get -> post/patch -> get". I understand that in this case we would read old values. And in the playbook new instance of the client is generated every time we use the module so that's why the module does work :)
I agree the CachedRestClient could be updated in a way that only inside certain methods in module (in my case get_role_uuids) CachedRestClient would be created from RestClient - but in this case type(self.client) changes from Client to RestClient and and this is no good :) Or I could updated module so that "run" function accepts Client and then in each function inside the module I create RestClient or CachedRestClient - depends on what is needed. 
But since the module is working properly I can also leave it as it is.

justinc1 (approved):
Thank you

PR #83, opened 2023-02-01
Ci experiment

justinc1:
I want to use ansible-community/ansible-test-gh-action action to have less boilerplate code for CI (prepare venv etc).
I created a new workflow file for this - I do not want to break existing CI.

I used parts of @shoriminimoe [!matrix-testing](https://github.com/ScaleComputing/HyperCoreAnsibleCollection/tree/matrix-testing) branch:
- sanity and units tests do not need self-hosted runner (thank you for noticing @shoriminimoe)
- use matrix test

The tests are run in docker containers, as this is default. Before this we didn't run sanit tests on ansible 2.14. Now we do run them, and some sanity tests required small code fixes. Similar, we didn't run unit tests with ansible 2.9, now we do, again small fixes were needed in the tests.

Sanity tests are run only for latest 3 ansible versions (2.12 to 2.14), on only one python version. I do not see need to cover combinations. In addition, pylint in ansible2.9 complained about whitespaces that later versions do not. This was most likely a bug in particular pylint/ansible-test version. Best to not even run this one.
`
Units test are run for all ansible/python version that collection is supposed to support. This should detect if we use some very recent python language feature, or import library that is not available yet in say python 3.8.

Integration tests - the PR does not yet touch them. Will be done later, I want first to get feedback on this one, and hopefully also merge it.


.github/workflows/ansible-test.yml
```diff
@@ -0,0 +1,40 @@
...
+      matrix: ${{ fromJSON(needs.job1.outputs.version_matrix) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Perform sanity testing with ansible-test
+        uses: ansible-community/ansible-test-gh-action@release/v1
+        with:
+          ansible-core-version: ${{ matrix.ansible }}
+          target-python-version: ${{ matrix.python }}
```
shoriminimoe, line 39: :bulb: idea

The docker containers used in the `ansible-test-gh-action` include all of the required python versions. We could remove this explicit option to test against all of the python versions automatically.

Unfortunately the `ansible-test-gh-action` action doesn't have a way to say "test only these python versions" or "don't test these ones". We could probably use the `pre-test-cmd` to remove the python versions we don't want to include.

.github/workflows/ansible-test.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Ansible test
+on:
+  push:
```
shoriminimoe, line 4: :crystal_ball: looking ahead

We will want to limit which kinds of pushes trigger integration tests when the time comes. It isn't necessary to run them on every push of _every_ branch. We probably only need on the default branch and at the end of pull requests.

shoriminimoe (approved):
@justinc1 this looks good. Great work. I left a couple comments with ideas/suggestions.

Looking forward, I think it would be good to make the integration test step dependent on the units test step. That way we don't run needless integration tests if we know there are other problems. That won't work very well if the units and sanity steps are combined as a single task as you have it now. Making the integration tests dependent on the sanity tests is unnecessary in my mind.

.github/workflows/ansible-test.yml
justinc1: Agreed.

.github/workflows/ansible-test.yml
justinc1: I will try to llet ansible-test auto-select ansible versions. If this works well, no need to complicate with `target-python-version: ${{ matrix.python }}`

justinc1: Hm. ansible 2.14 tries to run units test on python 2.7, and this fails. I do not want to uninstall whatever is in container just to prevent this test combination. Easier to explicitly list what we want.

.github/workflows/ansible-test.yml
```diff
@@ -0,0 +1,40 @@
...
+    outputs:
+      version_matrix: ${{ steps.set-matrix.outputs.version_matrix }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - id: set-matrix
+        name: set version_matrix output
+        # run: echo "::set-output name=matrix::{\"node\":[10, 12, 14]}"
+        run: python .github/workflows/print_ansible_test_matrix.py >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
justinc1: I will use a plain json/yaml file.

.github/workflows/ansible-test.yml
PolonaM: I think adding "config.yml" in tests folder with content:
modules:
  python_requires: ">=3.8"
should skip unit tests with python2.7. When I run ansible-test units I get: "WARNING: Skipping unit tests on Python 2.7 because it is not supported by this collection. Supported Python versions are: 3.8, 3.9, 3.10, 3.11"

shoriminimoe: Are you sure that is because of the config or is it because you don't have those versions installed on your machine? The test action has all of the python versions installed.

shoriminimoe: I was hoping the `pre-test-cmd` option would allow us to remove python versions from the docker containers, but unfortunately it runs before the containers are created. :disappointed:

shoriminimoe: > Hm. ansible 2.14 tries to run units test on python 2.7, and this fails. I do not want to uninstall whatever is in container just to prevent this test combination. Easier to explicitly list what we want.

I tried a handful of ways to prevent testing older python versions with the `ansible-test-gh-action` action. It seems like it simply can't be done without setting the python version explicitly as you are doing.

  I think that is what ultimately deterred me from using the action in [my testing branch](https://github.com/shoriminimoe/HyperCoreAnsibleCollection/tree/matrix-testing). I found it easier to use the `setup-python` action with the versions I care about then use `ansible-test --venv` to test all of the python versions automatically.

PR #84, opened 2023-02-01
Add virtual_disk_info module

justinc1:
Module sample output is:
```
  sample:
    name: demo-virtual-disk
    block_size: 1048576
    size: 1073741824
    replication_factor: 2
    uuid: 7983b298-c37a-4c99-8dfe-b2952e81b092
```

`size` gets value from `capacityBytes` in API response. API field `totalAllocationBytes` is ignored, until it gets clear why it is equal to `capacityBytes`.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -91,6 +91,7 @@ jobs:
     runs-on: [self-hosted2]
     container: python:3.10-slim-buster
     env:
+      DEBIAN_FRONTEND: noninteractive
       ANSIBLE_COLLECTIONS_PATH: /work-dir
       # vm_import|vm_export - they need SMB server
       # dns_config|time_server - NTP cannot be reconfigured if DNS is invalid
@@ -105,7 +106,7 @@ jobs:
       - run: apt update
       - run: apt install -y git make
       - run: pip install ansible-core==2.13.1
-      - run: apt install -y genisoimage
+      - run: apt install -y genisoimage qemu-utils
       - run: apt install -y jq
       - run: mkdir -p $WORKDIR
       - run: cp -a ./  $WORKDIR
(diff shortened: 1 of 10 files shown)


justinc1:
Raw API response is like:
```
TASK [Show particular disk info] ***************************************************************
ok: [localhost] => 
  result.record[0]:
    blockSize: 1048576
    capacityBytes: 1073741824
    name: ci-test-virtual-disk-1.qcow2
    replicationFactor: 2
    totalAllocationBytes: 1073741824
    uuid: 55eec75d-008b-4f1f-af2d-054fbbb4a933
```

Can you agree about proposed virtual_disk_info module output:
```
    - name: List a single virtual disk
      scale_computing.hypercore.virtual_disk_info:
        name: "{{ image_filename }}"
      register: result
    - ansible.builtin.assert:
        that:
          - result is succeeded
          - result is changed
          - result.records | length == 1
          - result.records.0.block_size == 1048576
          - result.records.0.capacity_bytes == 1073741824
          - result.records.0.name == "{{ image_filename }}"
          - result.records.0.replication_factor == 2
          - result.records.0.total_allocation_bytes == 1073741824
          - result.records.0.uuid | len == 36
```

domendobnikar:
`capacityBytes` is Total capacity, in bytes.
`totalAllocationBytes` is Storage space in use, in bytes.
`blockSize` is Size of individual blocks (the smallest unit of measurement for VSDs) on the drive, in bytes.

Do we want to rename any of these, or are these standard names that users can understand?
`total allocation` vs `capacity` can be a bit misleading. Maybe => `allocated capacity` and `total capacity` or `allocated size` and `total size` to be more in line with vm_disk module ... we use `size` there instead of `capacity` I think.

justinc1:
I decided to use `size` for `capacityBytes`, to be similar to vm_disk module.

The `totalAllocationBytes` - for now I will just remove it from output. I see totalAllocationBytes==capacityBytes, this is not what I would expect.

justinc1:
Assigned also @PolonaM , because HyperCore version check will be used also in cluster_name module.

domendobnikar (approved):
Looks good, Hypercore version check will come in handy.

PR #92, opened 2023-02-02
Us03 remote support tunnels

PolonaM:
Added support_tunnel and support_tunnel_info module.

### DIFF
--- a/plugins/module_utils/support_tunnel.py
+++ b/plugins/module_utils/support_tunnel.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 __metaclass__ = type
 
 from ..module_utils.utils import PayloadMapper
-from .errors import UnexpectedAPIResponse
 from ansible.module_utils.basic import AnsibleModule
 from ..module_utils.client import Client
 from ..module_utils.typed_classes import TypedSupportTunnelToAnsible
--- a/plugins/module_utils/support_tunnel.py
+++ b/plugins/module_utils/support_tunnel.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2023, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+from __future__ import annotations
+
+__metaclass__ = type
+
+from ..module_utils.utils import PayloadMapper
+from .errors import UnexpectedAPIResponse
+from ansible.module_utils.basic import AnsibleModule
+from ..module_utils.client import Client
+from ..module_utils.typed_classes import TypedSupportTunnelToAnsible
+from typing import Any
+
+
+class SupportTunnel(PayloadMapper):
+    def __init__(self, tunnel_open: str):
+        self.tunnel_open = tunnel_open
+
+    @classmethod
+    def from_ansible(cls, ansible_data: Any) -> Any:
+        pass
+
+    @classmethod
+    def from_hypercore(cls, hypercore_data: dict[str, str]) -> SupportTunnel:
+        # There is no None check since get_record is not used (support_tunnel's api behaves different)
+        return cls(
+            tunnel_open=hypercore_data["tunnelOpen"],
+        )
+
+    def to_hypercore(self) -> Any:
+        pass
+
+    def to_ansible(self) -> TypedSupportTunnelToAnsible:
+        return dict(
+            tunnel_open=self.tunnel_open,
+        )
+
+    def __eq__(
+        self, other: object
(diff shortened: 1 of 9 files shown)


domendobnikar:
Do you know if we can have multiple tunnels opened at the same time?
If we can, there could be problems with info_module.
Otherwise looks good. 👍

domendobnikar (commented):
Do you know if we can have multiple tunnels opened at the same time?
If we can, there could be problems with info_module.
Otherwise looks good. 👍

PolonaM:
> Do you know if we can have multiple tunnels opened at the same time? If we can, there could be problems with info_module. Otherwise looks good. +1

Thanks for the comment! Only one support tunnel can be opened, so no problems with the info :)

plugins/module_utils/support_tunnel.py
```diff
@@ -0,0 +1,72 @@
...
+        This method is used only in tests.
+        """
+        if not isinstance(other, SupportTunnel):
+            return NotImplemented
+        return self.tunnel_open == other.tunnel_open
+
+    @classmethod
+    def check_tunnel_status(cls, client: Client) -> SupportTunnel:
+        response = client.get("/support-api/check")  # type: ignore
+        if response.status == 200:  # get lets 404 through
```
justinc1, line 56: When do we get 404? If there is typo in URL, or also is some other case?

plugins/modules/support_tunnel.py
```diff
@@ -0,0 +1,146 @@
...
+
+- name: Close support tunnel
+  scale_computing.hypercore.support_tunnel:
+    state: absent
+"""
+
+RETURN = r"""
+record:
+  description:
+    - User record.
```
justinc1, line 56: copy-paste error

plugins/modules/support_tunnel.py
```diff
@@ -0,0 +1,146 @@
...
+  returned: success
+  type: dict
+  sample:
+    - tunnel_open: 4422
```
justinc1, line 60: Why not `code: 4422`? Then you return same name as the input.

plugins/modules/support_tunnel_info.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+
+RETURN = r"""
+record:
+  description:
+    - Support tunnel status.
+  returned: success
+  type: dict
+  sample:
+    - tunnel_open: 4422
```
justinc1, line 38: `code: 4422` if you agree.

tests/integration/targets/support_tunnel/tasks/main.yml
```diff
@@ -0,0 +1,89 @@
+---
+- environment:
+    SC_HOST: "{{ sc_host }}"
+    SC_USERNAME: "{{ sc_username }}"
+    SC_PASSWORD: "{{ sc_password }}"
+    SC_TIMEOUT: "{{ sc_timeout }}"
+
+  block:
+    - name: Cleanup
```
justinc1, line 9: It looks like returned `tunnel_open` is sometimes int, sometimes boolean. Can we return the same datatype always? We can also say it is not number, but a string, and than "" is a special value meaning tunnel is not closed, and non-empty value means tunnel is opened. Class `SupportTunnel` has `__init__(tunnel_open: str` too.

justinc1 (changes requested):
Thank you. If we treat the code/tunnel_open value as string, things are sort of simpler. We must not know it is number anyway. Simple in sense, return value from support_tunnel is not sometimes int, sometimes false.

plugins/module_utils/support_tunnel.py
domendobnikar: image
In Client we let 200 and 404 through.
Do we need to fix this?

tests/integration/targets/support_tunnel/tasks/main.yml
domendobnikar: We could also add `open: bool` as one of the output parameters. That way we make it easier for testing (True/False) and less ambiguous for user. If you both agree @justinc1 @PolonaM .

plugins/module_utils/support_tunnel.py
PolonaM: @justinc1 In case of support_tunnel only 200 is acceptable, but now after testing all the bad cases I can see only 400 is returned which is already handled in the client, so I will remove the "404 check" from the code. 
@domendobnikar client.py is ok for now

plugins/modules/support_tunnel.py
PolonaM: Updated

tests/integration/targets/support_tunnel/tasks/main.yml
PolonaM: @domendobnikar I agree with the proposal, so - we have code: int and open: bool

justinc1 (approved):

PR #94, opened 2023-02-05
use more runners for CI jobs

justinc1:
Added 10+ runners with label `self-hosted2. Now try to use them.

CI configuration is (except for passwords) stored in `CI_CONFIG_HC_IP50` variable. It contains a whole yml document.


anazobec (approved):
Approved! :+1:

PR #96, opened 2023-02-06
US01 - add type annotation

PolonaM:
Type annotation added to user and user_info module

### DIFF
--- a/changelogs/fragments/user_modules_type_annotation.yml
+++ b/changelogs/fragments/user_modules_type_annotation.yml
@@ -1,3 +1,0 @@
----
-minor_changes:
-  - Added type annotation to user modules. (https://github.com/ScaleComputing/HyperCoreAnsibleCollection/pull/96)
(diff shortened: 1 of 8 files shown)


changelogs/fragments/user_modules_type_annotation.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - Added type annotation to user modules. (https://github.com/ScaleComputing/HyperCoreAnsibleCollection/pull/96)
```
domendobnikar, line 3: This is probably not needed.
Document fragments should be used in case of an "operational/functional" change.
Examples: Bugfix, adding/removing/altering input or output, new module, new functionality.

domendobnikar (approved):
Looks good 👍

changelogs/fragments/user_modules_type_annotation.yml
PolonaM: I agree, I removed the fragment

PR #97, opened 2023-02-06
Speedup CI jobs

justinc1:
a_vm and git_issues were the slowest. https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4101531667 - each took about 30 min.

vm test is splitted into 3 tests. The 02_shutdown is tested twice, with nice ACPI shutdown, and with force shutdown.

`git_issues` - I disabled that one. Plan is to run it once a day only. Will be in separate PR, because those slow tests are annoying...

Force shutdown - maybe we want to test this one on daily basis too?


.github/workflows/ci-test.yml
```diff
@@ -92,7 +92,10 @@ jobs:
     container: python:3.10-slim-buster
     env:
       ANSIBLE_COLLECTIONS_PATH: /work-dir
-      SKIP_TESTS: "vm_import|vm_export|dns_config|time_server|time_zone"
+      # vm_import|vm_export - they need SMB server
+      # dns_config|time_server - NTP cannot be reconfigured if DNS is invalid
+      # git_issues - slow, do not run on each push. TODO - run them only once a day
```
domendobnikar, line 97: **GIt_issues**
How about pull_request only?
Either when pull request is created or when it's approved.
Should be run at least before merge to check if any existing code is broken.

tests/integration/targets/vm/tasks/02_shutdown.yml
```diff
@@ -1,44 +1,43 @@
 ---
 # This is a part of the vm module; testing shutdown logic.
-# The main vm test was devided into smaller tests for sake of synchronous testing.
+# The main vm test was divided into smaller tests for sake of synchronous testing.
+# Question: do we want to test nice ACPI shutdown, or force shutdown, or both?
+# We should test correct force shutdown too.
```
domendobnikar, line 5: Maybe we can test both shutdowns here in file 02.yml
After that we can divide VM test into two separate folders "shutdown" and "main-operations" something like that.
We can divide even further if necessary. But this should already create two ~10 min tests instead of one that takes ~20 min.

domendobnikar (approved):

.github/workflows/ci-test.yml
domendobnikar: Dispatch was the agreed upon solution.
Github_issues can be manually triggered before the code is merged into main to prevent any existing code breaking.
👍

justinc1: will be run on push/merge to main, and manually (workflow-dispatch)

PR #98, opened 2023-02-07
Registration info

domendobnikar:
- Adding missing registration_info
- Adding documentation fragment for registration and registration info
- Adding default vars to registration integration test


justinc1 (approved):
Only one change for how to revert configuration to correct value after test.
Similar to https://github.com/ScaleComputing/HyperCoreAnsibleCollection/blob/main/tests/integration/targets/smtp/tasks/main.yml:
- It better if recovery is in `always` block.
- correct values for the cluster are in https://github.com/ScaleComputing/HyperCoreAnsibleCollection/settings/variables/actions/CI_CONFIG_HC_IP50 , under `...registration`

You can also just merge, and we modify test later.

PR #99, opened 2023-02-07
Add Us11 email alert modules (email_alert, email_alert_info)

anazobec:
Added email alert modules (email_alert, email_alert_info)

Examples usages:
```yaml
- name: Create a new Email Alert Recipient
  scale_computing.hypercore.email_alert:
    email: example@example.com
    state: present

- name: Update previously create Email Alert Recipient
  scale_computing.hypercore.email_alert:
    email: example@example.com
    email_new: new@example.com
    state: present
    
- name: Remove previously updated Email Alert Recipient
  scale_computing.hypercore.email_alert:
    email: new@example.com
    state: absent

- name: Send test email to an Email Alert Recipient
  scale_computing.hypercore.email_alert:
    email: recipient@example.com
    state: test

# ---- info
- name: Retrieve al recipients
  scale_computing.hypercore.email_alert_info:
  register: info
- ansible.builtin.debug:
    msg: "{{ info.records }}"
```

### DIFF
--- a/plugins/module_utils/email_alert.py
+++ b/plugins/module_utils/email_alert.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2023, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from .rest_client import RestClient
+from ..module_utils.utils import PayloadMapper, get_query
+from ..module_utils import errors
+
+
+class EmailAlert(PayloadMapper):
+    def __init__(
+        self,
+        uuid: str = None,
+        alert_tag_uuid: str = None,
+        email_address: str = None,
+        resend_delay: int = None,
+        silent_period: int = None,
+        latest_task_tag: {} = None,
+    ):
+        self.uuid = uuid
+        self.alert_tag_uuid = alert_tag_uuid
+        self.email_address = email_address
+        self.resend_delay = resend_delay
+        self.silent_period = silent_period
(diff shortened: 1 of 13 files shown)


plugins/module_utils/email_alert.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @classmethod
+    def from_ansible(cls, ansible_data: TypedEmailAlertFromAnsible):
+        return EmailAlert(
+            uuid=ansible_data["uuid"],
+            alert_tag_uuid=ansible_data["alert_tag_uuid"],
+            email_address=ansible_data["email_address"],
+            resend_delay=ansible_data["resend_delay"],
+            silent_period=ansible_data["silent_period"],
+            latest_task_tag=ansible_data["latest_task_tag"],
```
justinc1, line 49: I think you do not need to read from ansible fields that are not ansible module parameters.

plugins/modules/email_alert.py
```diff
@@ -0,0 +1,263 @@
...
+        after,
+        dict(before=before, after=after),
+    )  # changed, records, diff
+
+
+def send_test(module: AnsibleModule, rest_client: RestClient):
+    before = EmailAlert.get_state(rest_client)
+
+    if module.params["email_new"] is not None:
+        module.warn("Email Alert: parameter 'email_new' is not needed.")
```
justinc1, line 199: Can we ask ansible to prevent this - required_if or similar?

justinc1 (approved):
Changes can be implemented later.

plugins/modules/email_alert.py
anazobec: I was looking for something like ``not_required_if`` or ``hidden_if`` because, even when we use ``state=present`` the parameter ``email_new`` is not required. Unless we make the parameter ``email_new`` as required when ``state=present``, but when we don't want to update the current email, we just set ``email_new`` to ``""``?

PR #102, opened 2023-02-09
Add release workflow

shoriminimoe:
Add the release.yml workflow file for automatic releases. Releases are triggered by tagging a commit with a name matching `v*.*.*`.

The workflow follows these steps:
- Verify the release by comparing the tag name with the galaxy.yml version.
- Build the collection tar file to include as a release artifact.
- Generate release notes from CHANGELOG.rst
- Create a [github release](https://github.com/ScaleComputing/HyperCoreAnsibleCollection/releases) using the tag name
- Publish the release to ansible galaxy (**this step will fail until a valid api key is added to the repo secrets**)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,63 @@
+---
+name: Release and Publish
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup python 🐍
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install dependencies 👽
+        run: python -m pip install ansible-core yq rst-to-myst
+
+      - name: Verify release 🗒️
+        env:
+          TAG_NAME: ${{ github.ref_name }}
+        run: |
+          echo "::warning::TODO: confirm tests pass for this commit"
+          tag_version="${TAG_NAME##*v}"
+          version="$(yq -r .version galaxy.yml)"
+          if [[ ${tag_version} != ${version} ]]; then
+            echo "::error file=galaxy.yml::🔴 Tag version '${tag_version}' does not match galaxy.yml version '${version}'"
+            exit 1
+          fi
+          echo "::notice::🟢 Version number verified: '$version'"
+
+      - name: Build 🛠️
+        run: ansible-galaxy collection build -v --force .
+
+      - name: Prepare release notes 📓
+        env:
+          CHANGELOG_URL: "${{ format('{0}/{1}/blob/{2}/CHANGELOG.rst', github.server_url, github.repository, github.ref_name) }}"
+          REF_NAME: "${{ github.ref_name }}"
+        run: |
+          rst2myst stream --no-sphinx CHANGELOG.rst | csplit - "/^## ${REF_NAME}/" '/^## v.*/'
+          mv xx01 release_notes.md
+          echo "See also: [CHANGELOG](${CHANGELOG_URL}#${REF_NAME//./-})" >>release_notes.md
+
+      - name: Release 🛸
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            *.tar.gz
+          body_path: release_notes.md
+          fail_on_unmatched_files: true
+          generate_release_notes: true
+
+  publish:
+    needs: release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Publish to galaxy 🌌
+        uses: ansible/ansible-publish-action@v1.0.0
+        with:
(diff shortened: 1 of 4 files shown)


shoriminimoe:
Here are a few examples from my fork
- [workflow run](https://github.com/shoriminimoe/HyperCoreAnsibleCollection/actions/runs/4130246380)
- [release](https://github.com/shoriminimoe/HyperCoreAnsibleCollection/releases/tag/v1.1.1)

.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: python -m pip install ansible-core yq rst-to-myst
+
+      - name: Verify release 🗒️
+        env:
+          TAG_NAME: ${{ github.ref_name }}
+        run: |
+          echo "::warning::TODO: confirm tests pass for this commit"
+          tag_version="${TAG_NAME##*v}"
+          version="$(yq -r .version galaxy.yml)"
+          if [[ ${tag_version} != ${version} ]]; then
```
justinc1, line 28: We have extra "v" in tag, so `${tag_version} != v${version}`.

Seems in galaxy.yml, we will still have version without "v" - like here https://github.com/ansible-collections/azure/blob/v1.14.0/galaxy.yml.

.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Build 🛠️
+        run: ansible-galaxy collection build -v --force .
+
+      - name: Prepare release notes 📓
+        env:
+          CHANGELOG_URL: "${{ format('{0}/{1}/blob/{2}/CHANGELOG.rst', github.server_url, github.repository, github.ref_name) }}"
+          REF_NAME: "${{ github.ref_name }}"
+        run: |
+          rst2myst stream --no-sphinx CHANGELOG.rst | csplit - "/^## ${REF_NAME}/" '/^## v.*/'
```
justinc1, line 42: I think here we will need to use also https://github.com/ansible-community/antsibull-changelog. Maybe whole CHANGELOG.rst will not be part of git repo, but only generated from ./changelogs/

galaxy.yml
```diff
@@ -18,13 +18,10 @@ documentation: https://scalecomputing.github.io/HyperCoreAnsibleCollection-docs
 homepage: https://github.com/ScaleComputing/HyperCoreAnsibleCollection
 issues: https://github.com/ScaleComputing/HyperCoreAnsibleCollection/issues
 build_ignore:
-  - .idea
-  - .gitignore
+  - .*
```
justinc1, line 21: Nice catch :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            *.tar.gz
+          body_path: release_notes.md
+          fail_on_unmatched_files: true
+          generate_release_notes: true
+
+  publish:
+    needs: release
```
justinc1, line 56: Should we make this a manual/workflow-dispatch step?

.github/workflows/release.yml
shoriminimoe: You're right. I'm actually handling that on line 26 by removing the "v" (and everything preceding it) from the `TAG_NAME` environment variable: `tag_version="${TAG_NAME##*v}"`.

.github/workflows/release.yml
shoriminimoe: This part in the workflow assumes the changelog has already been updated using `antsibull-changelog`. I added that as a step to the release steps in the "[New release](https://github.com/ScaleComputing/HyperCoreAnsibleCollection/blob/main/.github/ISSUE_TEMPLATE/new-release.md)" issue template.

The process I threw together here is a bit disjointed. Here is how it should work right now:

1. Team decides there should be a release
2. Someone creates a "New release" issue for the version to be released and assigns it.
3. The assignee completes the release steps
    1. Update the version number in galaxy.yml e.g. `1.1.1`
    2. Update the changelog with `antsibull-changelog` e.g. `antsibull-changelog release --version 1.1.1`
    3. Tag the commit with the version number prefixed with 'v' e.g. `v1.1.1`
    4. Push the tag to github
4. Release workflow is triggered by the new tag.

I have a bunch of thoughts for improving this :laughing:

.github/workflows/release.yml
shoriminimoe: I think that is a good idea for the first iteration. It would be great to have this be automated at some point, but we need to sort out our security posture first.

.github/workflows/release.yml
justinc1: OK, thanks for explaining, this sounds nice.

.github/workflows/release.yml
justinc1: Merging. Thank you for helping with this

justinc1 (approved):

PR #103, opened 2023-02-09
Removing dict[Any, Any] from TypedDiff

domendobnikar:
- Removed unnecassary `dict[Any]` from `TypedDiff`.
- Small fix to an empty output in User module.


PolonaM:
I approve :)

PolonaM (approved):

PR #104, opened 2023-02-09
Fix Smtp bug on empty cluster

justinc1:
Bug is reproducible. Now fix.


justinc1 (commented):
One bug fixed, so I accept.

PR #106, opened 2023-02-09
Us07 OIDC config

domendobnikar:
Create, list, update and delete OIDC connect config.
Single playbook example to create Azure AD config and set up hypercore cluster.

- [x] OIDC action module
- [x] OIDC info module
- [x] Add documentation fragment
- [x] Add Example playbook (create Azure AD config and set up cluster)


tests/integration/integration_config.yml.template
```diff
@@ -47,5 +47,11 @@ sc_config:
       use_auth: false
       auth_user: ""
       auth_password: ""
+    oidc:
+      uuid: oidcconfig_uuid
+      client_id: "d2298ec0-0596-49d2-9554-840a2fe20603"
+      shared_secret: "~0lI5n-_8dNbWmrrqqTE1v1iBcfO4__jfx"
+      config_url: https://login.microsoftonline.com/76d4c62a-a9ca-4dc2-9187-e2cc4d9abe7f/v2.0/.well-known/openid-configuration
+      scopes: "openid+profile"
```
domendobnikar, line 55: Does this need to be added to github variables `CI_CONFIG_HC_IP50`?

justinc1: yes, this info should not be public.

plugins/modules/oidc_config.py
```diff
@@ -0,0 +1,153 @@
...
+                type="str",
+                required=True,
+            ),
+            shared_secret=dict(
+                type="str",
+                no_log=False,
+            ),
+            certificate=dict(
+                type="str",
+            ),
```
justinc1, line 131: `certificate` is used for client certificate authentication? If yes, please add `no_log: True`.

tests/unit/plugins/module_utils/test_oidc.py
```diff
@@ -0,0 +1,148 @@
...
+import sys
+
+import pytest
+
+from ansible_collections.scale_computing.hypercore.plugins.module_utils.oidc import (
+    Oidc,
+)
+
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (2, 7), reason="requires python2.7 or higher"
```
justinc1, line 19: import the "3.8" check from Polona

tests/unit/plugins/modules/test_oidc_config.py
```diff
@@ -0,0 +1,155 @@
...
+
+import pytest
+
+from ansible_collections.scale_computing.hypercore.plugins.modules import oidc_config
+from ansible_collections.scale_computing.hypercore.plugins.module_utils.oidc import (
+    Oidc,
+)
+
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (2, 7), reason="requires python2.7 or higher"
```
justinc1, line 15: 3.8

justinc1 (changes requested):
shared_secret should not be commited to code. Otherwise, nice work, thank you.

plugins/modules/oidc_config.py
```diff
@@ -0,0 +1,153 @@
...
+    unit: bool,
+) -> Tuple[bool, Union[TypedOidcToAnsible, None], TypedDiff]:
+    before = oidc_obj.to_ansible() if oidc_obj else None
+    oidc_obj_ansible = Oidc.from_ansible(module.params)
+    if oidc_obj is None:
+        task = oidc_obj_ansible.send_create_request(rest_client)
+    else:
+        task = oidc_obj_ansible.send_update_request(rest_client)
+    if not unit:
+        sleep(10)  # Wait for the cluster login (Avoid BAD GATEWAY response)
```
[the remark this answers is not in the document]
domendobnikar: No, wait_task requires a response from API and getting a response is the problem here - since we always get `BAD GATEWAY`.
When OIDC config is modified, user gets reconnected to the cluster and causes `BAD GATEWAY` response from the API during the reconnect process.
This is essentially a workaround - but the actual issue should be addressed on the API back-end in my opinion.

plugins/modules/oidc_config.py
domendobnikar: Yes I think it is. You can use it instead of the shared_secret. Nice catch.

tests/unit/plugins/module_utils/test_oidc.py
domendobnikar: 👍

tests/unit/plugins/modules/test_oidc_config.py
domendobnikar: 👍

tests/integration/integration_config.yml.template
domendobnikar: I changed  / removed all secret values from the examples, tests and configs.
But I cannot add them to GITHUB since I am not an admin. Do you have access @justinc1 ?

plugins/modules/oidc_config.py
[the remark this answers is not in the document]
justinc1: Then this is best we can do at the moment. We can create waiting-api issue for this.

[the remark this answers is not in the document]
justinc1: I acknowledge it is reasonable workaround. I accept it.
But magic value of 10 sec - for sure best value depends also on other tasks HyperCore might be executing, and also on how responsive is the OIDC provider.

I played a bit with API module, trying to call the endpoint 2-times in a row. Sometimes I get `502 Bad Gateway` (1st or 2nd call), and sometimes this just works - I get a taskTag back. But more about this later.

tests/integration/integration_config.yml.template
justinc1: I added them.

justinc1 (approved):
Thank you.

plugins/modules/oidc_config.py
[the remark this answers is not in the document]
domendobnikar: Not a good solution I agree.
Maybe the sleep value can be based on an already passed parameter timeout?
But even if this value shows to work 100% of the time, we would still get problems running other tasks in parallel to this.

PR #108, opened 2023-02-09
Fix missing timeout when Client is created

justinc1:
But default value None for timeout is the root problem. To easy to miss you need to set timeout.

Removed all default parameter values for Client.Client(). Normal modules should use Client.get_client() anyway, and now no-one will want to use Client.Client()


domendobnikar (approved):
Looks good.

PR #109, opened 2023-02-09
Ensure we have always a valid DNS resolver configured

justinc1:
At least the first entry should be a valid DNS resolver server.


justinc1:
dns_config integration test passed.
Merging now, as without this cluster is too often left with misconfigured DNS.
Later, many other tests fail, and cluster in not really usable.

PR #110, opened 2023-02-10
Fix dns_config on empty cluster.

anazobec:
dns_config module didn't work on an empty cluster, so a fix was required.
Also fixed dns config integration tests in 02_dns_config:
- missing ``register: result`` in 1st two tasks, causing tests to fail
- removing/commenting out the last task ``Restoring back to default`` due to test fail and finding that the restoration is done without this task.


tests/integration/targets/dns_config/tasks/02_dns_config.yml
```diff
@@ -17,6 +17,7 @@
       - 8.8.8.8
       - 1.0.0.1
     state: set
+  register: result
```
justinc1, line 20: Thank you for noticing this :)

justinc1 (approved):

PR #111, opened 2023-02-10
Email alert minor changes

anazobec:
- Removed unused ``from_ansible`` parameters from ``module_utils/email_alert``
- Added ``required_if`` to module to get rid of a warning message.


plugins/modules/email_alert.py
```diff
@@ -180,7 +180,7 @@ def delete_email_alert(module: AnsibleModule, rest_client: RestClient):
         dict(email_address=module.params["email"]), rest_client
     )
     if not delete_email:
-        module.warn("Email Alert: The email you're trying to remove, doesn't exist.")
+        module.warn("Email Alert: The email you're trying to remove doesn't exist.")
```
justinc1, line 183: No need to show warn if email was already removed. warn will signal to user something seems to be broken. But everything is OK.

plugins/modules/email_alert.py
```diff
@@ -244,6 +242,7 @@ def main() -> None:
                 required=True,
             ),
         ),
+        required_if=[("state", "present", ("email_new",))],
```
justinc1, line 245: I think this does not work as it should. "Test invalid parameter combination - email_new with state=absent" commit - was this desired result. E.g. fail if email_new is set and state==absent.

justinc1 (changes requested):
One more thing, module does return ALL configured emails. But usually modules return only part that they modify. So we should change "result.records[0].xyz" to "result.record.xyz". I missed that before :/.

That helps user to say create email, get (only) created email in result, and continue to work with created email. If user needs UUID of created email - it is the only email in the result, he does not need to search for it in list of all email.

plugins/modules/email_alert.py
anazobec: I agree, it doesn't work as it should. The problem is, that the ``email_new`` param has default value set to ``""``. If that default is removed, than it partially works as it should, but another problem arises - ``email_new`` is not required even when ``state=present``.

PR #112, opened 2023-02-13
Us04 cluster name

PolonaM:
Added cluster_name_info and cluster_name modules

### DIFF
--- a/plugins/module_utils/cluster.py
+++ b/plugins/module_utils/cluster.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2023, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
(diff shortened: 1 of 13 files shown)


justinc1 (changes requested):
I remember we needed to use raw api module to get cluster version before. That is a bit painful.

I suggest to rename modules to `cluster` and `cluster_info`. Then we can include cluster version into `cluster/cluster_info` output.

I added desired cluster name to github CI variables, and added commit (and rebased code), so that final "restore original name" uses this value.

The restore (`always: ...`) part of integration test fails on 10.5.11.50 - module changes name 2 times, then 3rd PATCH with api module fails with "500 internal error". I guess we need to wait on taskTag. This is the last commit I added ("WIP") - please amend it with whatever needs to be done.

PolonaM:
> 

Thanks for all the comments!
Cluster info module was renamed and icos_version added. 
I added wait_task to the Cluster_name module to avoid error 500. I preserved the module name since another module cluster_shutdown will be added to the collection.

PR #113, opened 2023-02-14
Time zone on empty cluster

anazobec:
Fixed time_zone module not working on an empty cluster in which case it now creates one instead of printing out a warning message.


justinc1 (approved):
Thank you. I will now try to run integ test.

PR #115, opened 2023-02-14
Time server on empty cluster

anazobec:
Fixed time_server module not working on an empty cluster in which case it now creates one instead of printing out a warning message.


justinc1 (approved):
Code looks ok. Just add also integration test.

PR #117, opened 2023-02-15
Add cluster_shutdown module

PolonaM:
Added cluster_shutdown module

### DIFF
--- a/plugins/modules/cluster_shutdown.py
+++ b/plugins/modules/cluster_shutdown.py
@@ -38,8 +38,16 @@ EXAMPLES = r"""
     force_shutdown: true
 """
 
-RETURN = r""" # WHAT TO RETURN?
+RETURN = r"""
 record:
+  description:
+    - Cluster that was shutdown.
+  returned: success
+  type: dict
+  sample:
+    - icos_version: 9.2.11.210763
+      name: PUB5
+      uuid: a5d9148c-37f7-4b43-843c-196751d3c050
 """
 
 from ansible.module_utils.basic import AnsibleModule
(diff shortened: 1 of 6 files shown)


plugins/modules/cluster_shutdown.py
```diff
@@ -0,0 +1,81 @@
...
+description:
+  - Shutdown cluster.
+version_added: 1.2.0
+extends_documentation_fragment:
+  - scale_computing.hypercore.cluster_instance
+seealso: []
+options:
+  force_shutdown:
+      description:
+        - Bypasses attempts to gracefully shut down all runnning VirDomains before system shutdown.
```
justinc1, line 27: This means all running VMs will be force shutdown? The host/node will not be force shutdown?

plugins/modules/cluster_shutdown.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+DOCUMENTATION = r"""
+module: cluster_shutdown
+
+author:
+  - Polona Mihalič (@PolonaM)
+short_description: Shutdown cluster.
+description:
+  - Shutdown cluster.
```
justinc1, line 19: Add to description if module shut down a single node, or all nodes in a cluster.

justinc1 (changes requested):
Currently only clarifications in `DOCUMENTATION` are needed.

plugins/modules/cluster_shutdown.py
PolonaM: In the documentation it says: "Shutdown the system.". And we have only one node in each cluster, so I can't really test if really all nodes do shut down. We should reconfigure our system to have one cluster with two nodes to be able to test this. But we can also trust the documentation :)

justinc1: Or we ask in slack, and then amend to description if needed.

justinc1 (approved):

PR #119, opened 2023-02-15
US13 - SSL certificate

domendobnikar:
As a user I would like to be able to post (upload) SSLcertificates to each node in cluster and verify (no API-but verify through https directly) .

- [ ] Certificate module
- [ ] Verify certificate

### DIFF
--- a/plugins/modules/certificate.py
+++ b/plugins/modules/certificate.py
@@ -1,0 +1,105 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2022, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+module: certificate
+
+author:
+  - Domen Dobnikar (@domen_dobnikar)
+short_description: Handles cluster SSL certificates.
+description:
+  - Can upload and verify SSL certificates.
+version_added: 1.1.0
+extends_documentation_fragment:
+  - scale_computing.hypercore.cluster_instance
+seealso: []
+options:
+  certificate:
+    description:
+      - Plain text of the X.509 PEM-encode certificate.
+    type: str
+    required: true
+  private_key:
+    description:
+      - Plain text of the RSA PEM-encoded private key.
+    type: str
+    required: true
+notes:
+  - C(check_mode) is not supported.
+"""
+
+EXAMPLES = r"""
+- name: Upload new certificate
+  scale_computing.hypercore.certificate:
+    private_key: this_is_the_private_key
+    certificate: this_is_the_certificate
+"""
+
+RETURN = r"""
+record:
+  description:
+    - Certificate record.
+  returned: success
+  type: dict
+  sample:
+    private_key: this_is_the_private_key
+    certificate: this_is_the_certificate
+"""
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ..module_utils import arguments, errors
+from ..module_utils.utils import is_changed
+from ..module_utils.client import Client
+from ..module_utils.rest_client import RestClient
+from ..module_utils.typed_classes import TypedDiff, TypedCertificateToAnsible
+from typing import Union, Tuple
+
+
+def ensure_present(
+    module: AnsibleModule,
+    rest_client: RestClient,
+) -> Tuple[bool, Union[TypedCertificateToAnsible, None], TypedDiff, bool]:
+    return is_changed(before, after), after, dict(before=before, after=after)
+
+
+def run(
+    module: AnsibleModule, rest_client: RestClient
+) -> Tuple[bool, Union[TypedCertificateToAnsible, None], TypedDiff, bool]:
+    return ensure_present(module, rest_client)
+
+
+def main() -> None:
+    module = AnsibleModule(
+        supports_check_mode=False,
+        argument_spec=dict(
+            arguments.get_spec("cluster_instance"),
+            private_key=dict(
+                type="str",
+                required=True,
+            ),
+            certificate=dict(
+                type="str",
+                required=True,
+            ),
+        ),
+    )
+
+    try:
+        client = Client.get_client(module.params["cluster_instance"])
+        rest_client = RestClient(client=client)
+        changed, record, diff = run(module, rest_client)
+        module.exit_json(changed=changed, record=record, diff=diff)
+    except errors.ScaleComputingError as e:
+        module.fail_json(msg=str(e))
+
+
+if __name__ == "__main__":
+    main()
--- a/plugins/module_utils/typed_classes.py
+++ b/plugins/module_utils/typed_classes.py
@@ -70,6 +70,10 @@ class TypedClusterToAnsible(TypedDict):
     icos_version: str
 
 
+class TypedCertificateToAnsible(TypedDict):
+    certificate: str
+
+
 # Ansible module return Diff dict {before:{} after:{}}
 class TypedDiff(TypedDict):
     before: Union[
@@ -77,6 +81,7 @@ class TypedDiff(TypedDict):
         TypedSupportTunnelToAnsible,
         TypedUserToAnsible,
         TypedClusterToAnsible,
+        TypedCertificateToAnsible,
         None,
     ]
     after: Union[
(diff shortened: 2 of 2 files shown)


plugins/modules/certificate.py
```diff
@@ -0,0 +1,151 @@
...
+    module = AnsibleModule(
+        supports_check_mode=False,
+        argument_spec=dict(
+            arguments.get_spec("cluster_instance"),
+            private_key=dict(
+                type="str",
+                required=True,
+                no_log=True,
+            ),
+            certificate=dict(
```
domendobnikar, line 134: Right now this is `NO_LOG=FALSE,` since we output certificate as string as Ansible output.
If this is something that should be hidden, we can change output to be something like `"Certificate was uploaded"` or `"Certificate is already in use"`.

plugins/modules/certificate.py
```diff
@@ -0,0 +1,151 @@
...
+        task = upload_cert(module, rest_client)
+        max_retries = 10
+        for ii in range(max_retries):
+            try:
+                TaskTag.wait_task(rest_client, task)
+                break
+            except (
+                errors.ScaleComputingError
+            ) as ex:  # ConnectionRefusedError not working
+                if str(ex) == "[Errno 111] Connection refused":
```
domendobnikar, line 109: Similar problem as with oidc module.
When we upload a certificate the cluster loses connection.

tests/integration/targets/certificate/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+        private_key: "{{ lookup('file', 'private_key_example.pem') }}"
+        certificate: "{{ lookup('file', 'certificate_example.crt') }}"
+      register: certificate_info
+    - ansible.builtin.assert:
+        that:
+          - certificate_info is succeeded
+          - certificate_info is changed
+          - certificate_info.record.certificate != default_cert_info
+
+    - name: Upload certificate to API with certificate module - IDEMPOTENCE
```
domendobnikar, line 57: If certificate parameter is marked as `NO_LOG=TRUE` , then this module is no longer idempotent

plugins/modules/certificate.py
justinc1: Agreed, `certificate` is public part (we need to hide only `private_key`).

plugins/modules/certificate.py
justinc1: You should replace `raise errors.ScaleComputingError(ex)` with just `raise` to re-raise exactly the same exception as the one you intercepted.

The logic to detect exact exception from message inside exception is a bit wrong - what if message for errno code is formatted differently (maybe even due to local language). Can you create issue to refactor this anti-pattern? 
We have is also in `Cluster.shutdown()` - https://github.com/ScaleComputing/HyperCoreAnsibleCollection/blob/main/plugins/module_utils/cluster.py#L77-L101. And maybe on more places.

The problematic line of code (why we do not get something like ConnectionRefusedError) could be in `Client._request()` https://github.com/ScaleComputing/HyperCoreAnsibleCollection/blob/main/plugins/module_utils/client.py#L118-L119

```
        except URLError as e:
            raise ScaleComputingError(e.reason)
```

plugins/modules/certificate.py
```diff
@@ -0,0 +1,156 @@
...
+def ensure_present(
+    module: AnsibleModule,
+    rest_client: RestClient,
+) -> Tuple[bool, Union[TypedCertificateToAnsible, None], TypedDiff]:
+    before: TypedCertificateToAnsible = dict(certificate=get_certificate(module))
+    # Don't upload if not necessary
+    if before["certificate"].replace("\n", "") == module.params["certificate"].replace(
+        "\n", ""
+    ):
+        return False, None, dict(before=None, after=None)
```
justinc1, line 101: Imagine you upload certA + keyB, later you figure out mistake, and try to fix by uploading certA + keyA.
Module will not be helpful, because it cannot know key is diffrent.

I think it is better to drop idempotence here, and always POST to server, always return changed=True.
We do same also with password for SMTP config, and password/secret for OIDC config.

plugins/modules/certificate.py
```diff
@@ -0,0 +1,156 @@
...
+"""
+
+RETURN = r"""
+record:
+  description:
+    - Certificate record.
+  returned: success
+  type: dict
+  sample:
+    certificate: this_is_the_certificate
```
justinc1, line 52: Change to
```
  sample:
    certificate: |
      -----BEGIN CERTIFICATE-----
      MIIGKzCCBBOgAwIBAgIUWaGzXfgSUuwwPJu3F2Q/Ru/O8JQwDQYJKoZIhvcNAQEL
      ...
      -----END CERTIFICATE-----
```

I never know that is the format of certificate (pem, der, p12, etc).

examples/certificate_signing_example.yml
```diff
@@ -0,0 +1,46 @@
+---
+- name: Example - how to generate private keys and sign certificates with openSSL
```
justinc1, line 2: `Example - how to generate private keys and sign certificates with openSSL and upload to HyperCore server`

justinc1 (changes requested):
I think we should drop idempotence for this module.

How we catch specific exception - we should have better code for this, but there are also other modules that catch the most generic ScaleComputingError exception, and than look at exception message. So this will be done later.

plugins/modules/certificate.py
domendobnikar: Hmm yes good point I forgot about formatting.
Will create a github issue to look into this and refactor code.

plugins/modules/certificate.py
domendobnikar: Yes good point this can definitely be an issue.
Will do.

plugins/modules/certificate.py
domendobnikar: I will add this to give more info. We do already have this info in the parameter description. 
` - File location containing the X.509 PEM-encode certificate.`

plugins/modules/certificate.py
domendobnikar: I changed code within `client.py` for `ConnectionRefused `and `ConnectionReset`. I also created a github issue to check every other module. 
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/issues/137

examples/certificate_signing_example.yml
domendobnikar: Changed.

plugins/module_utils/client.py
```diff
@@ -116,6 +116,19 @@ def _request(self, method, path, data=None, headers=None, timeout=None):
...
+                and e.args
+                and type(e.args[0]) == ConnectionRefusedError
+            ):
+                raise ConnectionRefusedError(e)
+            elif (
+                isinstance(e.args, tuple)
+                and e.args
+                and type(e.args[0]) == ConnectionResetError
+            ):
+                raise ConnectionResetError(e)
```
justinc1, line 131: Nice, should make exception catching much easier.

tests/integration/targets/certificate/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Upload certificate to API with certificate module
+      scale_computing.hypercore.certificate:
+        private_key: "{{ lookup('file', 'private_key_example.pem') }}"
+        certificate: "{{ lookup('file', 'certificate_example.crt') }}"
+      register: certificate_info
+    - ansible.builtin.assert:
+        that:
+          - certificate_info is succeeded
+          - certificate_info is changed
+          - certificate_info.record.certificate != default_cert_info
```
justinc1, line 55: We should check `certificate == certificate_example.crt`

tests/integration/targets/certificate/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Upload certificate to API with certificate module - IDEMPOTENCE
+      scale_computing.hypercore.certificate:
+        private_key: "{{ lookup('file', 'private_key_example.pem') }}"
+        certificate: "{{ lookup('file', 'certificate_example.crt') }}"
+      register: certificate_info_idempotence
+    - ansible.builtin.assert:
+        that:
+          - certificate_info_idempotence is succeeded
+          - certificate_info_idempotence is not changed
+          - certificate_info_idempotence.record.certificate != default_cert_info
```
justinc1, line 66: It should be changed - the private key must be uploaded. I think code already works this way.

We should check certificate == certificate_example.crt.

justinc1 (changes requested):
Minor changes/typos.

tests/integration/targets/certificate/tasks/main.yml
domendobnikar: Actually, forgot to remove idempotence check. Module is not idempotent anymore.

tests/integration/targets/certificate/tasks/main.yml
domendobnikar: Yes good idea

justinc1 (approved):
Thank yuo

PR #120, opened 2023-02-16
Add US12 - Syslog Server List module

anazobec:
Added modules for US12 - Syslog Server List for modifying endpoint ``/rest/v1/AlertSyslogTarget``:
- module_utils/syslog_server.py
- modules/syslog_server.py
- modules/syslog_server_info.py

Also added integration tests (**currently not working**).

### Example usage
```yaml
---
# ------------- Module syslog_server
- name: Create Syslog server
  scale_computing.hypercore.syslog_server:
    host: 10.5.11.222 # [ REQUIRED ]
    port: 514 # [ OPTIONAL ]
    protocol: udp # tcp --> [ OPTIONAL ]
    state: present # absent --> create, delete [ REQUIRED ]

- name: Update existing Syslog server
  scale_computing.hypercore.syslog_server:
    host: 10.5.11.222 # [ REQUIRED ]
    host_new: 1.2.3.4 # host IP to update the existing one with [ OPTIONAL ]
    port: 514 # can also be different than the existing  [ OPTIONAL ]
    protocol: udp # can also be different than the existing  [ OPTIONAL ]
    state: present # absent --> create, delete [ REQUIRED ]

# When deleting, only "host" param is required.
# Deleting by "host".
- name: Delete Syslog server
  scale_computing.hypercore.syslog_server:
    host: 10.5.11.222 # [ REQUIRED ]
    state: absent

# ------------- Module syslog_server_info
- name: List all Syslog servers
  scale_computing.hypercore.syslog_server_info:
  register: syslog_info_records
- ansible.builtin.debug:
    msg: "{{ syslog_info_records }}"
```


plugins/module_utils/syslog_server.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+
+    def to_hypercore(self) -> dict[Any, Any]:
+        return dict(
+            alertTagUUID=self.alert_tag_uuid,
+            host=self.host,
+            port=self.port,
+            protocol=self.protocol,
+            resendDelay=self.resend_delay,
+            silentPeriod=self.silent_period,
```
justinc1, line 78: `alertTagUUID`, `resendDelay`, `silentPeriod` - likely we should treat them as readonly HC3 API attributes.

Checking what webUI sends to HC3. websocket message was:
```
42["command",{"action":"update","type":"AlertSyslogTarget","commandID":"b08b8c33-d815-406c-84e7-7bfe53644aa4","data":{"host":"10.5.11.222","port":"514","protocol":"0","uuid":"b08b8c33-d815-406c-84e7-7bfe53644aa4"}},["authPassword","sharedSecret"]]
```
Confirmed, on update/patch those values are not sent to HC3.

plugins/modules/syslog_server.py
```diff
@@ -0,0 +1,320 @@
...
+    default: udp
+    choices: [ udp, tcp ]
+    description:
+      -  Network protocol used to send syslog alerts.
+  state:
+    type: str
+    choices: [ present, absent ]
+    description:
+      - The desired state of the syslog server on HyperCore API.
+      - If I(state=present) and C(host_new) wasn't provided,
```
justinc1, line 53: A bit hard to understand. Maybe:
```
      - If I(state=present) a new Syslog server will be added,
        or existing Syslog server with matching C(host) will be updated.
      - C(host) of existing Syslog server can be changed by specifying both C(host) and C(host_new).
```

plugins/modules/syslog_server.py
```diff
@@ -0,0 +1,320 @@
...
+
+    # Return if there are no changes
+    if payload == {
+        "host": before.get("host"),
+        "port": before.get("port"),
+        "protocol": before.get("protocol"),
+    }:
+        return False, before, dict(before=before, after=before)
+
+    # Otherwise, update with new email address
```
justinc1, line 211: `Otherwise, update with new parameters`

plugins/modules/syslog_server.py
```diff
@@ -0,0 +1,320 @@
...
+
+def run(module: AnsibleModule, rest_client: RestClient):
+    syslog_server = SyslogServer.get_by_host(
+        host=module.params["host"], rest_client=rest_client
+    )
+    state = module.params["state"]
+    if state == "present":
+        if syslog_server:
+            return update_syslog_server(module, rest_client)
+        return create_syslog_server(module, rest_client)
```
justinc1, line 256: Please pass the `syslog_server` to both `update_syslog_server` and `create_syslog_server`. They both call `SyslogServer.get_by_host` as the very first thing.

Looking a this a second time. What happens when we try to rename `host` to `host_new`? What reports integration test - does task 'name: Update Syslog server - HOST ONLY' pass?

plugins/modules/syslog_server.py
```diff
@@ -0,0 +1,320 @@
...
+    syslog_server = SyslogServer.get_by_host(
+        host=module.params["host"], rest_client=rest_client
+    )
+
+    protocol = get_protocol(module.params["protocol"])
+
+    # If that syslog server already exists, it will not be created again (no duplicates)
+    if syslog_server:
+        before = syslog_server.to_ansible()
+        return False, before, dict(before=before, after=before)
```
justinc1, line 143: It would be a bug if this happens. Returning `changed=False` would be incorrect, because `update_syslog_server` should be called, not `create_syslog_server`. Do not return `changed=False`.

You can just `assert syslog_server`.

tests/integration/targets/syslog_server/tasks/main.yml
```diff
@@ -0,0 +1,179 @@
...
+  block:
+    - name: Get starting hosts
+      scale_computing.hypercore.syslog_server_info:
+      register: info
+    - ansible.builtin.set_fact:
+        starting_servers: "{{ info.records }}"
+
+    # -----------------------------------------------------------
+
+    - name: Get current email_alert
```
justinc1, line 26: wrong name

tests/integration/targets/syslog_server/tasks/main.yml
```diff
@@ -0,0 +1,179 @@
...
+- environment:
+    SC_HOST: "{{ sc_host }}"
+    SC_USERNAME: "{{ sc_username }}"
+    SC_PASSWORD: "{{ sc_password }}"
+
+  vars:
+    default_port: 514
+    tcp_protocol: SYSLOG_PROTOCOL_TCP
+    udp_protocol: SYSLOG_PROTOCOL_UDP
+    create_syserver:
```
justinc1, line 11: typo in var name

tests/integration/targets/syslog_server_info/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+---
```
justinc1, line 1: We cannot run `syslog_server` and `syslog_server_info` in parallel. Move this file to be a `tests/integration/targets/syslog_server/tasks/01_syslog_server_info.yml`.

plugins/modules/syslog_server.py
```diff
@@ -0,0 +1,320 @@
...
+        if module.params["host_new"] is not None:
+            params += "host_new"
+        if module.params["port"] is not None:
+            params += "port"
+        if module.params["protocol"] is not None:
+            params += "protocol"
+
+    if params:
+        msg = ", ".join(params) + " can be used only if state==present"
+        module.fail_json(msg=msg)
```
justinc1, line 274: I think the second example in `EXAMPLES` does not comply with this.

We need to decide if `port` and `protocol` can be changed at same time as `name`.
I do not see why that could not be possible, or why it would not be logical.
But I want this to be consistent with other modules - them `vm` module in particular.

What `vm` does - looking at integration test - ups, we do not test `vm_name_new` at all. :/.
Relevant code is https://github.com/ScaleComputing/HyperCoreAnsibleCollection/blob/main/plugins/modules/vm.py#L378-L394. A guess - we have bug in vm module, try to rename old_vm twice, on first call it gets renamed, but on second call we do not find renamed VM, and instead create a new VM with vm_name_new, so we end with 2 VMs with name==vm_name_new. But an integration test is needed to confirm this.

I cannot tell for sure if you should support changing name at same time as `port`, `protocol`. I think we will want to allow this. Please wait with this change until it is clear. Or maybe you have a different opinion?

justinc1 (changes requested):

plugins/modules/syslog_server.py
anazobec: Yes, task 'name: Update Syslog server - HOST ONLY' passes, the idempotence of this task doesn't...

plugins/modules/syslog_server.py
anazobec: My opinion would be to allow changing ``port`` and ``protocol`` at the same time because, why not? Let's say a user has created a syslog server with a certain port and protocol, but then realized that the port and protocol he'd created were wrong and would like to change both of them. Would it not then be better for the user to be able to change both ``port`` and ``protocol`` in the same task (at the same time) than in two separate tasks? Just my view on this...

Also, I've used the ``user`` module to help with the implementation of this module.

plugins/modules/syslog_server.py
anazobec: I think I've figured out why exactly the idempotence of this task fails.
In task 'Update Syslog server - HOST ONLY', a syslog server host "0.0.0.0" is changed to "1.2.3.4" which makes server host "0.0.0.0" nonexistent. And then in the next task 'Update Syslog server - HOST ONLY - idempotence' it again tries to update syslog server host "0.0.0.0" to "1.2.3.4" but because that host ("0.0.0.0") doesn't exist anymore, it just creates one, hence why that test fails, when asserting ``result.changed == False`` and ``result.diff.before == result.diff.after``.

plugins/modules/syslog_server.py
justinc1: I agree with that - "to allow changing port and protocol at the same time".

Also VM module will be updated to allow changing name in same way as any other parameter we can change on VM. So we are consistent.

plugins/modules/syslog_server.py
```diff
@@ -0,0 +1,322 @@
...
+                choices=["present", "absent"],
+                required=True,
+            ),
+        ),
+    )
+
+    try:
+        client = Client.get_client(module.params["cluster_instance"])
+        rest_client = RestClient(client)
+        # validate_params(module)  # might actually not be needed
```
justinc1, line 314: fell free to throw validate_params out

justinc1 (approved):
I renamed and modified example a bit.

Thank you for work.

plugins/modules/syslog_server.py
anazobec: Resolved in #125 .

PR #121, opened 2023-02-17
fix 502 bad gateway problem with oidc_config module

justinc1:
Module will be easier to use if ignore or retry on 500/502 error is hidden inside module.
A small stress test is added to ensure we hit the API errors frequently.


domendobnikar (approved):
Looks good, thank you. 👍

PR #122, opened 2023-02-17
VM rename test and fix

justinc1:
Integration test is added to show that renaming VM a second time (idempotence test) does fail in current main.
Problem is (in most places) we search for VM by `vm_name` only, but we should by `vm_name` xor `vm_name_new`.
A helper function `VM.get_by_old_or_new_name` does this, and is used in VM module.

PR was draft:
- (partially done) unit tests need to be updated
- done - sanity tests
- todo - run all integration tests
- vm_nic, vm_disk, vm_params modules should also be updated.
  - I update vm_params module only, because only vm_params and vm modules have `vm_name_new` param
- maybe we can do extra cleanup, and stop using old `VM.get_by_name` - I'm not sure, need to check all affected modules.
  - I didn't even try this - unit test update would be painful 

But there is enough code for first review and comments

### DIFF
--- a/plugins/module_utils/vm.py
+++ b/plugins/module_utils/vm.py
@@ -359,6 +359,16 @@ class VM(PayloadMapper):
         return vm_from_hypercore
 
     @classmethod
+    def get_by_old_or_new_name(cls, ansible_dict, rest_client):
+        vm_old_name = VM.get_by_name(ansible_dict, rest_client)
+        vm_new_name = VM.get_by_name(ansible_dict, rest_client, name_field="vm_name_new") if ansible_dict.get("vm_name_new") is not None else None
+        if vm_old_name and vm_new_name:
+            # Having two candidate VMs is error, we cannot decide which VM to modify.
+            raise ScaleComputingError(f"More than one VM matches requirement vm_name=={ansible_dict['vm_name']} or vm_name_new=={ansible_dict['vm_name_new']}")
+        vm = vm_old_name or vm_new_name
+        return vm
+
+    @classmethod
     def import_vm(cls, rest_client, ansible_dict):
         cloud_init = cls.create_cloud_init_payload(ansible_dict)
         data = cls.create_export_or_import_vm_payload(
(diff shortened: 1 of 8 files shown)


plugins/module_utils/vm.py
```diff
@@ -358,6 +358,19 @@ def get_by_name(
         vm_from_hypercore = cls.from_hypercore(hypercore_dict, rest_client)
         return vm_from_hypercore
 
+    @classmethod
+    def get_by_old_or_new_name(cls, ansible_dict, rest_client):
+        vm_old_name = VM.get_by_name(ansible_dict, rest_client)
+        if ansible_dict["vm_name_new"] is None:
```
domendobnikar, line 364: `vm_name_new = ansible_dict["vm_name_new"]`
Can probably remove this `if ansible_dict["vm_name_new"] is None:` and just do the `else`.
The code `vm = vm_old_name or vm_new_name` already checks for `None`.

justinc1: You are right, this wil be better, thank you. And VM.get_by_name does return None if nothing is found, so the 
`vm_name_new = VM.get_by_name() if ... else None` reduces to just `vm_name_new = VM.get_by_name()`.

justinc1:
Unit test problems with TestManageVMDisks - I didn't check why I have problems with TestManageVMNics, but not with TestManageVMDisks. At first look, those two tests are not that much different.

domendobnikar (approved):
Looks good, I will add a backlog task / issue for VM code cleanup.
- Unit test cleanup
- Get method cleanup

plugins/module_utils/vm.py
```diff
@@ -358,6 +357,26 @@ def get_by_name(
         vm_from_hypercore = cls.from_hypercore(hypercore_dict, rest_client)
         return vm_from_hypercore
 
+    @classmethod
+    def get_by_old_or_new_name(cls, ansible_dict, rest_client, must_exist=False):
```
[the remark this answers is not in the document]
justinc1: We can have a talk about this. I'm a bit confused by those options (`find_by_name(must_exists=True)` vs `get_or_fail()`, etc).

I have a felling a more robust way would be not to search for VM by name all the time, but search for it by name 1st time, then access it by UUID.

Also, I have a felling having `VM.from_hypercore(vm_hypercore_dict)` accessing HC3 to get also node info and snapshot schedule info is not ideal. Passing two extra params like hypercode_nodes and hypercore_snapshot_schedules might make unit test easier to write - less magic mocking, more like dependency injection. But whole team should agree this is something worth to try, and maybe first on some smaller class, not on VM.

justinc1:
One more note, there is problem in VM.__str__() implementation. It will push branch problem-with-power-state-mapping, to show this. It seems to me that we store in VM.power_state both VM power_state as retrieved from HyperCore, and sometimes also power_state as set in ansible module. But ansible power_state is actually a command describing transition to HyperCore power_state, not HyperCore power_state (e.g. you can get into SHUTOFF state by nice shutdown or by force shutdown - two different actions, same end state).

I noticed this only because I added debug module.warn. Module was not misbehaving.

It would be a bit of extra work and testing, so I just added a small workaround to not throw this error to module users.

PR #123, opened 2023-02-20
Cluster_name update version check

PolonaM:
Corrected version requirement.
Also integration test updated so it doesn't fail in case if hcversion doesn't support cluster name patch method.


justinc1 (approved):
The recovery logic does recovery if `cluster_name` fails, and in same time it does relay on `cluster_name` module being properly implemented. At least it is supposed to return exactly correct error message.

I accept, since it does the job.

PR #124, opened 2023-02-20
Scheduled integration tests

justinc1:
We want to at least run integration tests periodically (if we cannot on every commit).
PR move integration test into a dedicated workflow, and the workflow is scheduled to be run at 6 AM.


PolonaM (approved):
Approved :)

PR #126, opened 2023-02-21
Make use_auth read-only (remove from create/update payload).

anazobec:
For the ``smtp`` module is better that the ``use_auth`` param is ``read-only``, so that was updated.


justinc1 (changes requested):
Current changes looks OK.

Please also remove `use_auth` in example, integration test, and in tests/integration/integration_config.yml.template.

justinc1 (approved):
Thank you

PR #127, opened 2023-02-21
Fix timeout error in case when a lot of VMs need to be shutdown

PolonaM:
Fixed timeout error in case when a lot of VMs need to be shutdown - the timeout error is ignored (but not the UnexpectedAPIResponse error). Also get request to rest/v1/Cluster is added to check when connection is refused, and only then the cluster_shutdown module successfully exits.


justinc1 (approved):
LGTM. I do not dear to test remotely ;)

PR #132, opened 2023-02-24
Test syslog_server_info with 0 and 1 syslog server present

justinc1:
Test
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4259825907/jobs/7412400903 failed, there was no syslog server configured.

One extra task for you: line at end of tests/integration/targets/syslog_server/tasks/01_syslog_server_info.yml:

```
          #- syslog_servers.records[0].protocol == "tcp"  # fix in module syslog_server_info needed here
```

Info module does not convert `SYSLOG_PROTOCOL_UDP` to `udp`, but it should.
The syslog_server_info module should return values that are same as those accepted by syslog_server module.
You want to be able to directly compare.
Please fix this is separate PR. You can create a new jira issue if you want.


anazobec (approved):
Looks good, thanks!
Also, will open a new PR for the requested syslog_server_info module fix.

PR #133, opened 2023-02-24
Do not run email_alert and smtp tests in parallel

justinc1:
Job
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4259825907/jobs/7412403066 failed with
TASK [email_alert : Send test email to an existing Email Alert Recipient] ****** fatal: [testhost]: FAILED! => {"changed": false, "msg": "Unexpected response - 400 b'{\"error\":\"SMTP unconfigured\"}'"}

The smtp test was run in parallel.
Avoid this by not running both at the same time - exclude smtp.


anazobec (approved):
Thanks!

PR #134, opened 2023-02-24
Fix syslog_syslog_server_info module output, update tests.

anazobec:
Fixed a minor issue pointed out in pull request #132.
> Info module does not convert ``SYSLOG_PROTOCOL_UDP`` to ``udp``, but it should.
> The syslog_server_info module should return values that are same as those accepted by syslog_server module.

### DIFF
--- a/plugins/module_utils/syslog_server.py
+++ b/plugins/module_utils/syslog_server.py
@@ -147,7 +147,7 @@ class SyslogServer(PayloadMapper):
         ]
 
         for s in state:
-            s["protocol"] = protocols[s["protocol"]]
+            s["protocol"] = protocols[s["protocol"]]  # type: ignore
 
         return state
 
--- a/plugins/module_utils/syslog_server.py
+++ b/plugins/module_utils/syslog_server.py
@@ -64,7 +64,7 @@ class SyslogServer(PayloadMapper):
             alert_tag_uuid=hypercore_data["alertTagUUID"],
             host=hypercore_data["host"],
             port=hypercore_data["port"],
-            protocol=hypercore_data["protocol"],
+            protocol=protocols[hypercore_data["protocol"]],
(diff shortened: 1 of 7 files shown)


plugins/module_utils/syslog_server.py
```diff
@@ -144,6 +146,9 @@ def get_state(
             )
         ]
 
+        for s in state:
+            s["protocol"] = protocols[s["protocol"]]  # type: ignore
+
```
justinc1, line 151: Please move this part of code into `from_hypercore()` method - this is the place to do mapping from hypercore representation into our internal representation.

tests/unit/plugins/module_utils/test_syslog_server.py
```diff
@@ -23,6 +23,9 @@
     reason=f"requires python{MIN_PYTHON_VERSION[0]}.{MIN_PYTHON_VERSION[1]} or higher",
 )
 
+PROTOCOL_TCP = "SYSLOG_PROTOCOL_TCP"
+ANSIBLE_TCP = "tcp"
```
justinc1, line 27: Variable names like `HYPERCORE_PROTOCOL_TCP` and `ANSIBLE_PROTOCOL_TCP` would be more explicit.

justinc1 (changes requested):
Just small changes. i feel a bit bad for being annoying. Hope is not to bad. :)

plugins/module_utils/syslog_server.py
anazobec: You shouldn't feel bad about it, that part of code was actually not good for my eyes either and knew there should be a better way to go about it, I just couldn't think of it at the moment. :)

anazobec:
Note: I have to fix unit tests. Missed that.

justinc1 (approved):
Thank you, merging.

PR #135, opened 2023-02-24
Version management

PolonaM:
Added modules:
- version_update_info (returns list of available updates)
- version_update (applies selected updated to the cluster)
- version_update_status_info (returns status of the latest applied update)

### DIFF
--- a/plugins/modules/version_update.py
+++ b/plugins/modules/version_update.py
@@ -48,7 +48,7 @@ record:
   type: dict
   sample:
     - uuid: 9.2.11.210763
-      description: 9.1.11 General Availability
+      description: 9.2.11 General Availability
       change_log: ...Please allow between 20-40 minutes per node for the update to complete...
       build_id: 210763
       major_version: 9
--- a/plugins/modules/version_update.py
+++ b/plugins/modules/version_update.py
@@ -43,13 +43,18 @@ EXAMPLES = r"""
 RETURN = r"""
 record:
   description:
-    - Cluster info.
+    - Version applied.
   returned: success
   type: dict
   sample:
-    - icos_version: 9.2.11.210763
-      name: cluster_name
-      uuid: a5d9148c-37f7-4b43-843c-196751d3c050
+    - uuid: 9.2.11.210763
+      description: 9.1.11 General Availability
+      change_log: ...Please allow between 20-40 minutes per node for the update to complete...
+      build_id: 210763
+      major_version: 9
+      minor_version: 2
+      revision: 11
+      timestamp: 0
 """
 
 from ansible.module_utils.basic import AnsibleModule
@@ -58,37 +63,33 @@ from ..module_utils import arguments, errors
 from ..module_utils.client import Client
 from ..module_utils.rest_client import RestClient
 from ..module_utils.cluster import Cluster
-from ..module_utils.hypercore_version import Update, UpdateStatus
-from ..module_utils.typed_classes import TypedClusterToAnsible, TypedDiff
-from typing import Tuple
-from time import sleep
+from ..module_utils.hypercore_version import Update
+from ..module_utils.typed_classes import TypedUpdateToAnsible, TypedDiff
+from typing import Tuple, Union, Dict, Any
 
 
 def run(
     module: AnsibleModule, rest_client: RestClient
-) -> Tuple[bool, TypedClusterToAnsible, TypedDiff]:
+) -> Tuple[bool, Union[TypedUpdateToAnsible, None], Dict[Any, Any]]:
     cluster = Cluster.get(rest_client)
     if cluster.icos_version == module.params["icos_version"]:
         return (
             False,
-            cluster.to_ansible(),
-            dict(before=cluster.to_ansible(), after=cluster.to_ansible()),
+            None,
+            dict(
+                before=dict(icos_version=cluster.icos_version),
+                after=dict(icos_version=cluster.icos_version),
+            ),
         )
-    Update.apply(rest_client, module.params["icos_version"])
-    while True:
-        sleep(30)
-        update_status = UpdateStatus.get(rest_client).status
-        if update_status == "IN PROGRESS":
-            pass
-        elif update_status == "COMPLETE":
-            break
-        else:
-            raise errors.UpdateNotSuccessfull
-    cluster_updated = Cluster.get(rest_client)
+    update = Update.get(rest_client, module.params["icos_version"], must_exist=True)
+    update.apply(rest_client)  # type: ignore
     return (
         True,
-        cluster_updated.to_ansible(),
-        dict(before=cluster.to_ansible(), after=cluster_updated.to_ansible()),
+        update.to_ansible(),  # type: ignore
+        dict(
+            before=dict(icos_version=cluster.icos_version),
+            after=dict(icos_version=update.uuid),  # type: ignore
+        ),
     )
 
 
--- a/plugins/modules/version_update.py
+++ b/plugins/modules/version_update.py
@@ -64,7 +64,7 @@ from ..module_utils.client import Client
 from ..module_utils.rest_client import RestClient
 from ..module_utils.cluster import Cluster
 from ..module_utils.hypercore_version import Update
-from ..module_utils.typed_classes import TypedUpdateToAnsible, TypedDiff
+from ..module_utils.typed_classes import TypedUpdateToAnsible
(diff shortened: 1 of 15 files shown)


plugins/modules/version_update.py
```diff
@@ -0,0 +1,114 @@
...
+    if cluster.icos_version == module.params["icos_version"]:
+        return (
+            False,
+            cluster.to_ansible(),
+            dict(before=cluster.to_ansible(), after=cluster.to_ansible()),
+        )
+    Update.apply(rest_client, module.params["icos_version"])
+    while True:
+        sleep(30)
+        update_status = UpdateStatus.get(rest_client).status
```
justinc1, line 80: I guess cluster/node API is restarted during update. We will need to handle errors like connection refused, HTTP invalid gateway, etc.

tests/integration/targets/version_update/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
+---
```
justinc1, line 1: We need to by default exclude this test from CI jobs - .github/workflows/integ-test.yml.

tests/integration/targets/version_update_status_info/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+    SC_TIMEOUT: "{{ sc_timeout }}"
+
+  block:
+    - name: Get status of the latest update applied
+      scale_computing.hypercore.version_update_status_info:
+      register: update_status
+    - ansible.builtin.assert:
+        that:
+          - update_status.record != []
+          - update_status.record.keys() | sort == ['from_build', 'percent', 'status', 'status_details', 'to_build', 'to_version']
```
justinc1, line 15: You can check values in 'percent', 'status', 'status_details' - they are "constant"

plugins/modules/version_update.py
```diff
@@ -0,0 +1,114 @@
...
+    module: AnsibleModule, rest_client: RestClient
+) -> Tuple[bool, TypedClusterToAnsible, TypedDiff]:
+    cluster = Cluster.get(rest_client)
+    if cluster.icos_version == module.params["icos_version"]:
+        return (
+            False,
+            cluster.to_ansible(),
+            dict(before=cluster.to_ansible(), after=cluster.to_ansible()),
+        )
+    Update.apply(rest_client, module.params["icos_version"])
```
justinc1, line 77: Try to update 9.2 to 9.1 - it will actually try to downgrade. Better to not try this. Or is this actually possible/supported?

justinc1 (commented):
One thing is not covered by module - the upgrade path. I mean if you are on 9.1.3, and wont to get to 9.2.20, then you need first upgrade to 9.1.18 (or similar).

I do not think this needs to be in first version of module.
I'm not sure if we need to implement this - if we do not implement, then we need to mention in documentation user needs to ensure update to provided version is valid/supported.

Maybe we also test/ask if attempt to do unsupported update can corrupt system.

If we need to somehow support upgrade to arbitrary version (e.g. multiple version), then this will be in this module, or in a additional role. We will need to have upgrade path list somewhere.

tests/integration/targets/version_update/tasks/main.yml
PolonaM: Yes you are right, my bad, I will exclude it from the tests

plugins/modules/version_update.py
PolonaM: It is only possible to update with the available updates in the list of updates. From what I have seen, available updates are always never than existing one. But nevertheless, shouldn't API take care of appropriate list of updates (or even downgrades)?

plugins/module_utils/errors.py
```diff
@@ -92,3 +92,9 @@ class VMInvalidParams(ScaleComputingError):
     def __init__(self):
         self.message = "Invalid set of parameters - strict affinity set to true and nodes not provided."
         super(VMInvalidParams, self).__init__(self.message)
+
+
+class UpdateNotSuccessfull(ScaleComputingError):
```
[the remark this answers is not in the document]
PolonaM: Ok, updated

tests/integration/targets/version_update_status_info/tasks/main.yml
PolonaM: True, will add the checks

plugins/modules/version_update.py
justinc1: You are right, if you check first that module input parameter `icos_version` is listed under available updates for this cluster, then you are safe.

plugins/module_utils/errors.py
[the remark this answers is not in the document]
justinc1: This part of code was removed.

plugins/modules/version_update.py
```diff
@@ -0,0 +1,115 @@
...
+
+RETURN = r"""
+record:
+  description:
+    - Version applied.
+  returned: success
+  type: dict
+  sample:
+    - uuid: 9.2.11.210763
+      description: 9.1.11 General Availability
```
justinc1, line 51: `9.2.11`, not `9.1.11`

plugins/modules/version_update_info.py
```diff
@@ -0,0 +1,86 @@
...
+
+RETURN = r"""
+records:
+  description:
+    - A list of updates that can be applied to this cluster.
+  returned: success
+  type: list
+  sample:
+    - uuid: 9.2.11.210763
+      description: 9.1.11 General Availability
```
justinc1, line 41: 9.2.11

tests/integration/targets/version_update/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+        icos_version: "{{ updates.records[0].uuid}}"
+      register: update
+    - ansible.builtin.assert:
+        that:
+          - update is changed
+          - update.record.uuid == updates.records[0].uuid
+          - update.diff.before.icos_version == cluster_initial.record.icos_version
+          - update.diff.after.icos_version == updates.records[0].uuid
+    
+    # - name: Get status of the latest update applied
```
justinc1, line 46: Now you check update is being applied? And do not expect it is already complete?

tests/integration/targets/version_update/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+          - cluster_update_config.magic_allow_string == "allow-cluster-update-test"
+    
+    - name: Get list of hc version updates
+      scale_computing.hypercore.version_update_info:
+      register: updates
+    
+    - name: Get initial version
+      scale_computing.hypercore.cluster_info:
+      register: cluster_initial
+
```
justinc1, line 23: Here you can also look at version_update_status_info - check no update is in progress.

justinc1 (changes requested):
Minor changes.

Question: will be role to upgrade cluster in a separate PR? I sort of prefer 2 smaller PRs. This one is practically finished.

plugins/modules/version_update.py
PolonaM: updated

plugins/modules/version_update_info.py
PolonaM: updated

tests/integration/targets/version_update/tasks/main.yml
PolonaM: Here I only confirm that the update was applied (if module exited without error), with module version_update_status_info we then check for the status (in progress, terminated, complete..)

tests/integration/targets/version_update/tasks/main.yml
PolonaM: added

tests/integration/targets/version_update/tasks/main.yml
PolonaM: Added check with version_update_status_info if update is "in progress"

PolonaM: @justinc1 I deleted this check, since it takes some time for the status to change and we don't really know how much time we have to wait - if we wait too little, than update_status will still be "COMPLETE" or "TERMINATING", if we wait too long then it can happen we skip the first change in the status (for example prepare_status="DOWNLOAD RPMS") and the update_status is then already "EXECUTING". Or it can even be "TERMINATING" if something went wrong early on. I think this check should only be implemented in a role.

justinc1 (approved):
Thank you.

PR #136, opened 2023-02-27
US1b - Add role for updating single-node systems.

anazobec:
Added a role for updating single-node systems, required for US1b.

### Example usage:
```yaml
---
- name: Update HC3 single-node system
  include_role:
    name: scale_computing.hypercore.version_update_single_node
  vars:
    scale_computing_hypercore_desired_version: 9.1.23.210897
```


roles/version_update_single_node/tasks/main.yml
```diff
@@ -0,0 +1,48 @@
...
+    validate_certs: false
+
+- name: Get all available VMs
+  scale_computing.hypercore.vm_info:
+  register: vms
+- ansible.builtin.debug:
+    var: vms
+
+- name: Shutdown VMs
+  scale_computing.hypercore.vm:
```
ddemlow: should this use vm_params instead?  (since only changing power state)

roles/version_update_single_node/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+        action: post
+        endpoint: "/rest/v1/Update/{{ item.uuid }}/apply"
+      when: available_updates
+      loop: "{{ available_updates }}"
+
+    - name: Start VMs
+      scale_computing.hypercore.vm:
+        vm_name: "{{ item.vm_name }}"
+        power_state: start
+      loop: "{{ vms }}"
```
justinc1, line 39: This will start all VMs. You need to start only those that were running initially.

roles/version_update_single_node/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+        action: get
+        endpoint: /rest/v1/Update
+      register: available_updates
+
+    - name: Update HC3
+      scale_computing.hypercore.api:
+        action: post
+        endpoint: "/rest/v1/Update/{{ item.uuid }}/apply"
+      when: available_updates
+      loop: "{{ available_updates }}"
```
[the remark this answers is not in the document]
anazobec: I am not really sure how it's those ``available_updates`` are sorted...
On our NUCs, there is only one update available, so it will show only ``9.1.23.210897``, on others, well, there are no updates available at all, because those NUCs are on the latest (not-yet-official) version ``9.2.x``, so the list is empty.

roles/version_update_single_node/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Get all available VMs
+      scale_computing.hypercore.vm_info:
+      register: vms
+    - ansible.builtin.debug:
+        var: vms
+
+    - name: Shutdown VMs
+      scale_computing.hypercore.vm:
+        vm_name: "{{ item.vm_name }}"
+        power_state: shutdown
```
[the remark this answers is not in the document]
anazobec: I resolved this by using ``vm_params`` module instead.

roles/version_update_single_node/tasks/main.yml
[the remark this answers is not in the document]
anazobec: Oh and if we have an update available, let's say ``9.1.23.210897``, which is also a uuid of an update, the update path/endpoint would be ``rest/v1/Update/9.1.23.210897/apply``.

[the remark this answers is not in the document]
justinc1: Yes, all true. We will decide how to filter available updates after we know could be there.

roles/version_update_single_node/tasks/update.yml
```diff
@@ -0,0 +1,48 @@
...
+- name: Show shutdown results
+  ansible.builtin.debug:
+    var: vm_shutdown_result
+
+# ---------------------UPDATE-----------------------
+
+# - name: Update HC3
+#   scale_computing.hypercore.api:
+#     action: post
+#     endpoint: "/rest/v1/Update/{{ item.uuid }}/apply"
```
justinc1, line 32: This will be replaced with call of module Polona is writing.
 
No other comments.

roles/version_update_single_node/tasks/update_status_check.yml
```diff
@@ -0,0 +1,31 @@
...
+- name: Check update status on HC3
+  block:
+    - name: Increment retry_count
+      ansible.builtin.set_fact:
+        retry_count: "{{ 0 if retry_count is undefined else retries | int + 1 }}"
+
+    - name: Pause before checking update status - checks will report FAILED-RETRYING until update COMPLETE/TERMINATED
+      ansible.builtin.wait_for:
+        timeout: 60
+      delegate_to: localhost
```
justinc1, line 11: After you trigger update, status is first "FAILED-RETRYING", then it switches to normal value.
Do I understand correctly?

Can you remove this task ? The next task also waits on scale_computing.hypercore.version_update_status_info.
I think you can. I also think you know why you can't :). If unsure, just add comment, that maybe it can be removed (testing is pain here).

tests/integration/targets/role_version_update_single_node/tasks/01_update.yml
```diff
@@ -0,0 +1,29 @@
...
+  register: api_version_after
+- ansible.builtin.debug:
+    var: api_version_after
+
+# --------------------------------------------------------------
+
+- name: Assert that api_version_after != api_version_before
+  ansible.builtin.assert:
+    that:
+      - api_version_after != api_version_before
```
justinc1, line 29: Good test :)

roles/version_update_single_node/tasks/update_status_check.yml
anazobec: As I am not entirely sure if by removing this task the role would still work, I've just added a comment to it (just to be careful, for the code works perfectly fine right now), as suggested.

roles/version_update_single_node/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+        icos_version: "{{ scale_computing_hypercore_desired_version }}"
+      register: update_result
+
+    - name: Check update status
+      include_tasks: update_status_check.yml
+
+    - name: Show update result
+      ansible.builtin.debug:
+        var: update_result
+
```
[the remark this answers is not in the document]
anazobec: Left that part of code as is

tests/integration/targets/role_version_update_single_node/tasks/01_update.yml
anazobec: Agreed :)

roles/version_update_single_node/tasks/main.yml
```diff
@@ -0,0 +1,90 @@
...
+  scale_computing.hypercore.version_update_info:
+  register: available_updates
+
+- name: Show available updates
+  ansible.builtin.debug:
+    var: available_updates
+
+- name: End host if no update available
+  ansible.builtin.meta: end_host
+  when: available_updates.records == []
```
justinc1, line 44: Those 3 lines are redundunt ("Check if desired update is available - end host if Not available" does cover this case too), right? I will remove them. Ping back if you disagree.

examples/version_update_single_node.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Update a HyperCore single-node system
+  hosts: localhost
+  connection: local
+  gather_facts: false
+
```
justinc1, line 6: I will add 'check: true' to example. If someone runs example, but forgets to change SC_HOST first - in such case it is safer to not actually start update.

justinc1 (approved):

roles/version_update_single_node/tasks/main.yml
anazobec: Agreed

examples/version_update_single_node.yml
anazobec: Agreed and thank you.

PR #139, opened 2023-03-01
cluster_config role

justinc1:
The PR adds a role for initial cluster configuration. Role primary usage is to configure an empty cluster.

A dict to fully configure cluster looks like:
```
    scale_computing_hypercore_cluster_config:
      name: cluster-a
      registration:
        company_name: New company a
        contact: John Smith a
        phone: 056789000
        email: john_smith_a@gmail.com
      dns:
        server_ips:
          - 1.1.1.1
          - 1.0.0.1
        search_domains:
          - subdomain_1.example.com
          - subdomain_2.example.com
      oidc:
        client_id: 12345a
        shared_secret: secret_stuff_a
        config_url: https://login.microsoftonline.com/76d4c62a-a9ca-4dc2-9187-e2cc4d9abe7f/v2.0/.well-known/openid-configuration
        scopes: openid+profile
      time_server: 2.pool.ntp.org
      time_zone: Europe/Ljubljana
      smtp:
        server: mail_a.example.com
        port: 25
        from_address: "ci-test-a@example.com"
      email_alerts:
        - user_a1@test.com
      syslog_servers:
        - host: 10.10.10.10
          port: 42
          protocol: tcp
``` 

As secondary usage, a partial cluster configuration is possible. User just needs to invoke role with stripped version of `scale_computing_hypercore_cluster_config` variable.

Some modules are able to unconfigure a system, and some cannot. For example, syslog_server module is able to remove configured syslog server, but OIDC cannot be removed once setup. I'm not sure if this role should allow unconfiguration of a system.  So I ask for advice here, what is best.

For first draft, role can remove syslog_servers, email alerts, dns server and search_domains. If user needs to remove other configuration, a corresponding module can be used.

On TODO list:
- [x] documentation - provided with argument spec, but we need to start using antsibull-docs for documentation rendering.
- [x] example
- [x] argument spec

### DIFF
--- a/examples/cluster_config.yml
+++ b/examples/cluster_config.yml
@@ -1,0 +1,53 @@
+---
+- name: Reconfigure HyperCore cluster settings
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  # We do not want to reconfigure cluster with bogus example values.
+  check_mode: true
+
+  vars:
+    cluster_configuration:
+      name: cluster-a
+      registration:
+        company_name: New company
+        contact: John Smit
+        phone: 056789000
+        email: john_smith@gmail.com
+      dns:
+        server_ips:
+          - 1.1.1.1
+          - 1.0.0.1
+        search_domains: []
+      # After OIDC is configured, it cannot be removed.
+      # Do not configure OIDC in example.
+      # oidc:
+      #  client_id: your_client_id
+      #  shared_secret: your_shared_secret
+      #  # certificate: plain_text_from_x509
+      #  config_url: https://login.microsoftonline.com/your_tenant_uuid/v2.0/.well-known/openid-configuration
+      #  scopes: openid+profile
+      time_server: pool.ntp.org
+      time_zone: Europe/Ljubljana
+      smtp:
+        server: smtp-relay.gmail.com
+        port: 25
+        use_ssl: false
+        from_address: cluster-a@example.com
+      email_alerts:
+        - admin@example.com
+      syslog_servers:
+        - host: 10.10.10.10
+          port: 514
+          protocol: udp
+
+  tasks:
+    - name: Show configuration
+      ansible.builtin.debug:
+        var: cluster_configuration
+
+    - name: Configure HyperCore cluster
+      include_role:
+        name: scale_computing.hypercore.cluster_config
+      vars:
+        scale_computing_hypercore_cluster_config: "{{ cluster_configuration }}"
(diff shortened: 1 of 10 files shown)


examples/cluster_config.yml
```diff
@@ -0,0 +1,53 @@
...
+          - 1.1.1.1
+          - 1.0.0.1
+        search_domains: []
+      # After OIDC is configured, it cannot be removed.
+      # Do not configure OIDC in example.
+      # oidc:
+      #  client_id: 12345a
+      #  shared_secret: secret_stuff_a
+      #  # certificate: plain_text_from_x509
+      #  config_url: https://login.microsoftonline.com/76d4c62a-a9ca-4dc2-9187-e2cc4d9abe7f/v2.0/.well-known/openid-configuration
```
domendobnikar, line 28: Do we want to show this URI?
It's a real value not an example.

roles/cluster_config/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+See [argument_specs.yml](../../roles/cluster_config/meta/argument_specs.yml).
+
+## Limitations
+
+- NA
+
+## Dependencies
+
+- NA
```
domendobnikar, line 21: Should we add information about dependencies on valid DNS / internet connection somewhere in here?
OIDC for example can't be configured in case of external AD without a valid connection.
Or is this something that is self explanatory?

examples/cluster_config.yml
domendobnikar, line 28: Do we want to commit this URI?
It's a real value not an example.

roles/cluster_config/README.md
domendobnikar, line 21: Should we add information about dependencies on valid DNS / internet connection somewhere in here?
OIDC for example can't be configured in case of external AD without a valid connection.
Or is this something that is self explanatory?
We could also add this information in the module if missing.

domendobnikar (approved):

examples/cluster_config.yml
justinc1: No, thank you, I missed this. I wanted to have some realistic example, but not real value.

roles/cluster_config/README.md
justinc1: I think this is not needed. About half of cluter configuration modules requires IP connectivity and DNS resolution (SMTP server, email alerts, syslog server, even NTP server).

This whole readme - I want to throw it away, and replace it with documentation generated from argument_spec.

PR #140, opened 2023-03-01
Replace 'email_address' to 'email' in records return.

anazobec:
Names of return values didn't match the names of input parameters.
- renamed ``email_address`` to ``email``


justinc1 (approved):
Thank you.

PR #141, opened 2023-03-01
Handle case when code is already in use

PolonaM:
In case when support tunnel code is already in use, module now raises and error.


justinc1 (approved):
Thank you.

PR #142, opened 2023-03-01
Replace 'time_zone' with 'zone' in record return.

anazobec:
Names of return values didn't match the names of input parameters.
- renamed ``time_zone`` to ``zone``


justinc1 (approved):
Thank you again.

PR #147, opened 2023-03-01
use_auth is not a parameter for smpt module

justinc1:
Remove it from docs. Also mention module expects both or none of `auth_user`, `auth_password`.


anazobec (approved):
Thanks! Approved and merging.

PR #152, opened 2023-03-06
Fix valid time zone list

justinc1:
We got an authoritative time zone list. The PR includes the list into time_zone module.


anazobec (approved):
Thank you.

PR #153, opened 2023-03-06
M3 - Virtual disk - US02 - US03 - Create / Delete / List

domendobnikar:
As a user I want to be able to upload foreign virtual disks to the cluster from local file (list formats allowed - qcow2, vmdk, vhd):

Important: all conversion will be done by API

As a user I want to be able to list and delete uploaded virtual disks on cluster 

- [x] Create
- [x] List
- [x] Delete
- [x] Add document fragment

Attach will be done in a separate PR.


plugins/modules/virtual_disk.py
```diff
@@ -0,0 +1,197 @@
...
+from ..module_utils.state import State
+
+from ..module_utils.hypercore_version import (
+    HyperCoreVersion,
+)
+
+HYPERCORE_VERSION_REQUIREMENTS = ">=9.2.10"
+
+
+def read_disk_file(module: AnsibleModule) -> Tuple[bytes, int]:
```
domendobnikar, line 97: Add:
`Try .... catch : File not found`

plugins/modules/virtual_disk.py
```diff
@@ -0,0 +1,197 @@
...
+    virtual_disk_obj: Optional[VirtualDisk],
+) -> Tuple[bool, Union[TypedVirtualDiskToAnsible, None], TypedDiff]:
+    before = None
+    after = None
+    if not virtual_disk_obj:
+        return False, after, dict(before=before, after=after)
+    else:
+        before = virtual_disk_obj.to_ansible()
+        virtual_disk_obj.send_delete_request(rest_client)
+        # No wait_task needed; Upload not returning task ID.
```
domendobnikar, line 139: Remove comment, implement `wait_task()`.
Function itself should handle empty task ID.

plugins/modules/virtual_disk.py
```diff
@@ -0,0 +1,197 @@
...
+            ),
+        ),
+        required_if=[("state", "present", ("file_location",), False)],
+    )
+
+    try:
+        client = Client.get_client(module.params["cluster_instance"])
+        rest_client = RestClient(client)
+        hcversion = HyperCoreVersion(rest_client)
+        if not hcversion.verify(HYPERCORE_VERSION_REQUIREMENTS):
```
domendobnikar, line 187: Maybe add a `version_check()` function instead of this.
Make this a one function call.

plugins/module_utils/rest_client.py
```diff
@@ -98,33 +98,29 @@ def delete_record(self, endpoint, check_mode, timeout=None) -> TypedTaskTag:
...
+        endpoint: str,
+        payload: Union[dict[Any, Any], None],
+        check_mode: bool,
+        query: Union[dict[Any, Any], None] = None,
+        timeout: int = None,
+        binary_data: bytes = None,
         headers=None,
-    ):
+    ) -> TypedTaskTag:
         # Method put doesn't support check mode # IT ACTUALLY DOES
```
domendobnikar, line 109: Remove this comment, no point.

plugins/module_utils/virtual_disk.py
```diff
@@ -93,14 +98,50 @@ def __eq__(self, other: object) -> bool:
...
+            "/rest/v1/VirtualDisk/upload",
+            payload=None,
+            binary_data=file_content,
+            check_mode=False,
+            query=dict(filename=file_name, filesize=file_size),
+        )
+
+    def send_delete_request(self, rest_client: RestClient) -> TypedTaskTag:
+        return rest_client.delete_record(
+            f"/rest/v1/VirtualDisk/{self.uuid}", check_mode=False
```
domendobnikar, line 146: `if uuid:`

tests/integration/targets/virtual_disk/tasks/main.yml
```diff
@@ -0,0 +1,214 @@
...
+      scale_computing.hypercore.virtual_disk:
+        state: present
+        file_location: xlab-ci-test-VD.qcow2
+        file_name: xlab-ci-test-VD.qcow2
+      register: uploaded_virtual_disk
+    - ansible.builtin.assert:
+        that:
+          - uploaded_virtual_disk is succeeded
+          - uploaded_virtual_disk is changed
+          - uploaded_virtual_disk.record.name == "xlab-ci-test-VD.qcow2"
```
justinc1, line 39: Check also `uploaded_virtual_disk.record.keys() == [...]` is as expected, and also their values.

tests/integration/targets/virtual_disk/tasks/main.yml
```diff
@@ -0,0 +1,214 @@
...
+      scale_computing.hypercore.virtual_disk:
+        state: present
+        file_location: xlab-ci-test-VD.qcow2
+        file_name: xlab-ci-test-VD.qcow2
+      register: uploaded_virtual_disk
+    - ansible.builtin.assert:
+        that:
+          - uploaded_virtual_disk is succeeded
+          - uploaded_virtual_disk is not changed
+          - uploaded_virtual_disk.record.name == "xlab-ci-test-VD.qcow2"
```
justinc1, line 63: Check `uploaded_virtual_disk.record.uuid` and other attributes did not change.

justinc1 (changes requested):
Minor changes. Renaming is just great thing to generate work...

justinc1:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4466336327/ - jobs that failed here are not related to this PR (I believe). CI passes.

plugins/module_utils/virtual_disk.py
```diff
@@ -43,21 +45,20 @@ def from_ansible(cls, ansible_data: TypedVirtualDiskFromAnsible) -> VirtualDisk:
...
+            return cls(
+                name=hypercore_data["name"],
+                uuid=hypercore_data["uuid"],
+                block_size=hypercore_data["blockSize"],
+                size=hypercore_data["capacityBytes"],
+                # allocated_size=hypercore_data["totalAllocationBytes"],
+                replication_factor=hypercore_data["replicationFactor"],
+            )
+        except KeyError as e:
+            raise errors.MissingValueHypercore(e)
```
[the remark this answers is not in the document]
domendobnikar: It can potentially happen if virtual disk is missing parameter data (corrupted data on cluster) and is then returned with missing keys. 
We had some examples in the past broken with VMs.
This is just to give a better error LOG when fixing bugs.
@just
Do you think there is a potential downside to this?

plugins/module_utils/virtual_disk.py
```diff
@@ -93,14 +94,62 @@ def __eq__(self, other: object) -> bool:
...
+        rest_client: RestClient, file_content: bytes, file_size: int, file_name: str
+    ) -> TypedTaskTag:
+        if file_content is None or file_size is None or not file_name:
+            raise errors.ScaleComputingError(
+                "Missing some virtual disk file values inside upload request."
+            )
+        return rest_client.put_record(
+            "/rest/v1/VirtualDisk/upload",
+            payload=None,
+            binary_data=file_content,
```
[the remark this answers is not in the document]
domendobnikar: I will check and change if possible.

plugins/modules/virtual_disk.py
```diff
@@ -0,0 +1,212 @@
...
+version_added: 1.2.0
+extends_documentation_fragment:
+  - scale_computing.hypercore.cluster_instance
+seealso:
+  - module: scale_computing.hypercore.virtual_disk_info
+options:
+  file_location:
+    type: str
+    description: Location of the disk file.
+  file_name:
```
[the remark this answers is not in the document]
domendobnikar: Is name/source enough information for user to understand they need to specify a file instead of something else?
But from consistency POV I agree we should have the same parameter names where possible.

plugins/module_utils/virtual_disk.py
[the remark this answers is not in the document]
domendobnikar: Added a TODO comment to change after ISO bug is fixed.

plugins/module_utils/rest_client.py
```diff
@@ -112,28 +113,30 @@ def delete_record(
...
                 data=payload,
                 query=query,
                 timeout=timeout,
                 binary_data=binary_data,
                 headers=headers,
-            ).json
+            )
+            try:
+                response: TypedTaskTag = tt.json
+            except Exception:
```
justinc1, line 136: No. I pushed 'TEMP TEMP ..." commit just to show the `.json` part is problem. But we cannot catch arbitrary exception and essentially ignore it.

Who is calling this code? Maybe we need to let caller decide if conversion to json is possible? Or just catch JsonDecodeError in virtual_disk.py? Can you in this case be sure all data was uploaded - I think yes.

justinc1 (changes requested):
Avoid catching arbitrary exception

plugins/module_utils/rest_client.py
domendobnikar: Oh I thought this was a part of temporary ISO bug fix, so I just copy pasted your example.

plugins/module_utils/client.py
```diff
@@ -48,7 +48,7 @@ def json(self) -> Any:
             try:
                 self._json = json.loads(self.data)
             except ValueError:
-                raise ScaleComputingError(
+                raise ValueError(
```
domendobnikar, line 51: This was the actual problem ... We should really remove all `raise ScaleComputingError` from client and rest_client.

domendobnikar:
Integration tests:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4502644952

plugins/modules/iso.py
```diff
@@ -125,20 +126,27 @@ def ensure_present(module, rest_client):
...
+                    "Accept": "application/json",
+                    "Content-Length": file_size,
+                },
+            )
+    except FileNotFoundError:
+        raise errors.ScaleComputingError(
+            f"ISO file {module.params['source']} not found."
         )
+    except (json.JSONDecodeError, ValueError):
+        pass  # ISO API endpoint returns binary content.
```
domendobnikar, line 149: We can `pass` here, since the status code from response is acceptable.
Unacceptable status codes are raised in the client ... in case ISO upload was unsuccessful.

justinc1 (changes requested):
I suggest to use specific exception, not the generic ValueError. The palce where we erally decide that thia needs to be ignored - it is a few calls away, and ValueError could be triggered by other code, not just this `json.loads`.

plugins/module_utils/rest_client.py
```diff
@@ -133,7 +134,10 @@ def put_record(
             ).json
         except TimeoutError as e:
             raise errors.ScaleComputingError(f"Request timed out: {e}")
-
+        except (json.JSONDecodeError, json.decoder.JSONDecodeError) as e:
+            raise json.JSONDecodeError(e.msg, e.doc, e.pos)
+        except ValueError as e:
+            raise ValueError(e)
```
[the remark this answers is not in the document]
domendobnikar: https://github.com/ScaleComputing/HyperCoreAnsibleCollection/issues/165
I though you were also getting `JSONDecodeError`, I wasn't able to reproduce, I was only getting the `ValueError`, but I added it just in case since it was mentioned in the issue.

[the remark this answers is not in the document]
justinc1: Aha. I was thinking `JSONDecodeError` is being raised - it surprised me it is ValueError.

plugins/modules/iso.py
domendobnikar: I think so yes. This is the only module where we "pass" these json exceptions.
Other modules can handle them on their own if needed.
We should probably add these to other modules where we use `put`, but raise `ScaleError ` instead of pass.

justinc1 (approved):
Thank you. LGTM. I will run integ test manually.

PR #155, opened 2023-03-07
Add examples run to workflow.

anazobec:
Added workflows to run examples from ``examples/`` directory.
Examples are run in check mode.

### DIFF
--- a/.github/workflows/integ-test.yml
+++ b/.github/workflows/integ-test.yml
@@ -120,12 +120,11 @@ jobs:
           echo "matrix=$(
             ls -r $WORKDIR/examples | jq -R -s -c 'split("\n")[:-1]'
           )" >> $GITHUB_OUTPUT
-    if: "${{ github.event.schedule }}"
+    if: "${{ github.event.inputs || github.event.schedule }}"
 
   examples-run:
     needs:
       - examples-matrix
-      # - units-test
     runs-on: [ self-hosted2 ]
     container: python:3.10-slim-buster
     env:
@@ -134,9 +133,6 @@ jobs:
       fail-fast: false
       matrix:
         ansible: [ 2.13.0 ]
-        # python: [3.11]
-        # test_name: [user_info]
-        test_name: ${{ fromJson(needs.integ-matrix.outputs.matrix) }}
         example_name: ${{ fromJson(needs.examples-matrix.outputs.matrix) }}
     steps:
       - name: Checkout
@@ -148,12 +144,8 @@ jobs:
       - run: pip install ansible-core~=${{ matrix.ansible }}
       - run: mkdir -p $WORKDIR
       - run: cp -a ./  $WORKDIR
-      # We have ansible.cfg "for testing" in git repo
-      # (it is excluded in galaxy.yml, so it is not part of collection artifact)
-      # But it does affect ansible-galaxy and ansible-test commands.
       - run: cd $WORKDIR && ansible-galaxy collection install community.crypto
       - run: cd $WORKDIR && ansible-galaxy collection list
-        # TODO - make integration_config.yml a reusable artifact/output of integration-prepare-env
       - run: |
           cd $WORKDIR
           echo "${{ vars.CI_CONFIG_HC_IP50 }}" > tests/integration/integration_config.yml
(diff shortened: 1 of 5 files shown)


.github/workflows/integ-test.yml
```diff
@@ -113,6 +123,7 @@ jobs:
         # python: [3.11]
         # test_name: [user_info]
         test_name: ${{ fromJson(needs.integ-matrix.outputs.matrix) }}
+        example_name: ${{ fromJson(needs.integration-prepare-env.outputs.matrix) }}
     steps:
```
justinc1, line 148: The matrix does generate matrix product. Now could would try to run a commands for each "test_name, example_name" combination. You need to copy both job definitions, and give them unique names (integ_bla -> example_bla).

.github/workflows/integ-test.yml
```diff
@@ -65,6 +67,13 @@ jobs:
           ls -al tests/integration/integration_config.yml
       - run: cd $WORKDIR && ansible-playbook tests/integration/prepare/prepare_iso.yml
       - run: cd $WORKDIR && ansible-playbook tests/integration/prepare/prepare_vm.yml
+
+      - id: set-matrix
+        shell: bash
+        run: |-
+          echo "matrix=$(
+            ls -r $WORKDIR/examples | jq -R -s -c 'split("\n")[:-1]'
+          )" >> $GITHUB_OUTPUT
```
justinc1: This will be run only after `prepare_iso.yml`, `prepare_vm.yml`. Then, relatively fast test-examples will need to wait on those slow preparations. Move this to separate job. One job just to generate matrix, another job just to run `ansible-playbook --check examples/$example_name`.

justinc1 (changes requested):

.github/workflows/integ-test.yml
anazobec: To generate the matrix, I made a separate job ``examples-matrix`` as suggested. Then I also create a separate job to run those examples - ``examples-run``.

.github/workflows/integ-test.yml
anazobec: I created a separate job for running examples from the computed matrix - ``examples-run``.

justinc1 (approved):
LGTM. I will now run CI job manuall, and we see. :) Thank you.

justinc1:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4363093118 - does not yet work. Will add a few more comments to code.

.github/workflows/integ-test.yml
```diff
@@ -98,6 +98,72 @@ jobs:
...
+      - run: mkdir -p $WORKDIR
+      - run: cp -a ./  $WORKDIR
+
+      - id: set-matrix
+        shell: bash
+        run: |-
+          echo "matrix=$(
+            ls -r $WORKDIR/examples | jq -R -s -c 'split("\n")[:-1]'
+          )" >> $GITHUB_OUTPUT
+    if: "${{ github.event.schedule }}"
```
justinc1, line 123: I see now - I cannot run this manually (workflow-dispatch).

https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-event-information
https://docs.github.com/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_dispatch

I think this would already allow manual or scheduled start (if `github.event.inputs.integ_tests_include` is None or "" or [], then `github.event.inputs` is a dict, not None) .
```
    if: "${{ github.event.inputs || github.event.schedule }}"

```

.github/workflows/integ-test.yml
```diff
@@ -98,6 +98,72 @@ jobs:
...
+    container: python:3.10-slim-buster
+    env:
+      DEBIAN_FRONTEND: noninteractive
+    strategy:
+      fail-fast: false
+      matrix:
+        ansible: [ 2.13.0 ]
+        # python: [3.11]
+        # test_name: [user_info]
+        test_name: ${{ fromJson(needs.integ-matrix.outputs.matrix) }}
```
justinc1, line 139: You need to drop `test_name` from here.

justinc1 (changes requested):
small changes

justinc1 (approved):

PR #157, opened 2023-03-08
Hypercore version

domendobnikar:
Make it a one function call.
This might have to be implemented in multiple modules in the future.
Easier to maintain and fix if necessary ... all in one place.


PolonaM (approved):

justinc1 (approved):
Thank you.

PR #164, opened 2023-03-13
Ci docker image

justinc1:
We will use a custom docker image to run CI tests. Before we started with `python:3.10-slim-buster`, and need to run `apt install make git jq ...`, `pip install *.requirements`, and `pip install ansible` each time. All those tools will be part of our docker image.

This is example job with new docker image - https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4384232780/jobs/7675462776. It took about 20 sec.

The image contains latest ansible-2.13. I expect we will need at some point built more images (for ansbible 2.13, 2.14, etc). This is reason `ARG ANSIBLE_CORE_VERSION=2.13.0` parameter in Dockerfile - it allows to override ANSIBLE_CORE_VERSION outside of Dockerfile.

The docker image is stored under my personal quay.io account. Everyone can be pulled it, but login is needed for pushing new image/tag. For time being, only I will be able to push. We will switch to arbitrary docker repo as soon as Scale Computing says they have some other preference.

Since image is build locally, it is build with docker cache disabled (`--no-cache` option). Reason - when I will try to build new image after say 1 month, I do not want dockerd to use cached result from `apt update && apt install x y z` command.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -2,83 +2,73 @@ name: CI tests
 on:
   push:
 env:
-  WORKDIR: /work-dir/ansible_collections/scale_computing/hypercore
+  # ansible-test needs special directory structure.
+  # WORKDIR is a subdir of GITHUB_WORKSPACE
+  WORKDIR: work-dir/ansible_collections/scale_computing/hypercore
 jobs:
   mypy:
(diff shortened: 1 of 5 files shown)


anazobec (approved):
Looks good, thanks. :+1:

PR #167, opened 2023-03-13
Role for checking if local time is within required time interval

PolonaM:


roles/check_local_time/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    SC_TIMEOUT: "{{ sc_timeout }}"
+
+  block:
+    - name: Get Time Zone
+      scale_computing.hypercore.time_zone_info:
+      register: time_zone
+
+    - name: Check if local time is in time interval (run local_time.py)
+      ansible.builtin.script:
+        executable: python
```
justinc1, line 15: Potential problem, on Ubuntu we might have python3 , but no python.
Test
```
docker run -it -w /code -v $PWD:/code ubuntu:latest
apt-get update
apt-get install -y python3
  # we have python3 now

apt-get install -y python
  # no such thing, but in mentions you might want to install python-is-python3
```

Alternatives:
- use python3 - but you must test it will be accessible on major distributions
- refactor code to be invoked as a module (a module we want to have for internal use only, hm)
- you can check which python3 or python is on path (`which python3` or `which python`). I sort of prefer that one.

tests/integration/targets/role_check_local_time/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: Retrieve info about Time Zone
+      scale_computing.hypercore.time_zone_info:
+      register: time_zone
+
+    - name: Check if local time meets required time interval
+      ansible.builtin.include_role:
+        name: scale_computing.hypercore.check_local_time
+      vars:
+        time_zone: time_zone.record.zone
+        time_interval: 11:00-1:30 # Sometimes will pass, sometimes will not
```
justinc1, line 18: An optional param (default is "UTC now") containing seconds since epoch to explicitly set current time. it will allow testing whole thing.

PolonaM: @justinc1 In unit tests I did some refactoring, so I was able to write the tests. In integration test I solved this problem with using ansible_date_time variable (it's not perfect, but I think is better than implementing optional param just for testing). But now I have different problem - how to register the result from the role - I would like to assert the success or fail message but have no idea how to do that - any tip for that? :)

roles/check_local_time/README.md
```diff
@@ -0,0 +1,30 @@
+# cluster_config
```
justinc1, line 1: Change title

roles/check_local_time/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Role check_local_time can be used to:
+- check if local time in provided time zone is located within required time interval
```
justinc1, line 4: Maybe add extra line or two:

This can be used to ensure certain tasks/playbooks are executed only during allowed time interval.
For example, HyperCore version upgrade might be allowed only outside of business hours.

roles/check_local_time/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+- environment:
+    SC_HOST: "{{ sc_host }}"
+    SC_USERNAME: "{{ sc_username }}"
+    SC_PASSWORD: "{{ sc_password }}"
+    SC_TIMEOUT: "{{ sc_timeout }}"
```
justinc1, line 6: Those lines should be used only in integration tests. There, ansible-test will load variables from `integration-config.yml` - `sc_host` etc. Our modules are cofigured via environ `SC_HOST`, so first part of each integration test is that boilerplate code.

But in the actual roles and modules - do not do this. The `SC_HOST` environ must be set, or module needs to be invoked with `cluster_instance`.

tests/integration/targets/role_check_local_time/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+    SC_PASSWORD: "{{ sc_password }}"
+    SC_TIMEOUT: "{{ sc_timeout }}"
+
+  block:
+    - name: Check that local time meets required time interval
+      ansible.builtin.include_role:
+        name: scale_computing.hypercore.check_local_time
+      vars:
+        time_zone: "{{ ansible_date_time.tz }}"
+        time_interval: "{{ ansible_date_time.hour }}:00-{{ ansible_date_time.hour }}:59"
```
justinc1, line 14: It would be easier to test if current time would be additional input variable. It could be optional, with default value == ansible_date_time.

The current test - it can fail when run near x:00 or x:59 or x:01. Mention this in comment.

Test also demonstrates usage. It would be easier to use if role would just fail if current time is not within `time_interval`. Otherwise, every user will need to copy-paste 2 or 3 tasks - easy to add also a typo (variable names).

justinc1 (commented):
I suggest that role just fails if current time is not withing `time_interval`.

All other comments are minor.

roles/check_local_time/README.md
PolonaM: Added extra lines

roles/check_local_time/tasks/main.yml
PolonaM: removed from role

tests/integration/targets/role_check_local_time/tasks/main.yml
PolonaM: @justinc1 Role does fail if current time is not within time_interval, that's why apply: ignore errors is used in integration test. But still I needed something to assert the error (so that random error is not passed through) - now I am using ansible fact. But when user will use this role, these additional tasks will not be needed.

roles/check_local_time/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: Assert that local time is in time interval
+  ansible.builtin.assert:
+    fail_msg: "Local time for time zone {{ time_zone }} is not in required time interval {{ time_interval }}"
+    success_msg: "Local time for time zone {{ time_zone }} is in required time interval {{ time_interval }}"
+    that:
+      - local_time_output.stdout_lines[0] == "True"
+  register: result
+
+- name: Set fact to use in tests
+  ansible.builtin.set_fact: local_time_msg="{{ result.msg }}"
```
justinc1, line 33: No need to change this. But

Returned facts are global variables. Now is role user would already  use variable local_time_msg, then role would replace it. Some role authors then use extremely long fact names - like rolename_varname. Very safe, super ugly.

PolonaM:
@justinc1 User doesn't need to do anything, just use the role, provide the time_zone and time_interval and role will pass or fail. Role itself contains assertion that will pass or fail. Facts are there only for the tests

tests/integration/targets/role_check_local_time/tasks/main.yml
PolonaM: @justinc1 in case you missed the comment

justinc1:
I missed that, thank you for explaining. Perfect.

justinc1 (approved):

PR #168, opened 2023-03-14
Fix typo in update_status_check, fix force shutdown.

anazobec:
There was a typo in ``roles/version_update_single_node/update_status_check.yml``, so that had to be fixed.

In ``roles/version_update_single_node/shutdown_vms.yml`` shutting down VMs was done by using "stop" power state (force shutdwon) right away. This was changed, so that we first try shutting down VMs usinng the "shutdown" power state (regular shutdown) and then, if that fails, we use "stop" power state to forcibly shut down the rest of the VMs.


justinc1 (approved):
Thank you

PR #169, opened 2023-03-14
Document VM shutdown options

justinc1:
Extend VM shutdown options documentation. I do not want to duplicate what is already documented in Scale docs, but this one is just easy to miss. We ended using force shutdown during node upgrade - bad. So we need a bit more documentation.


PolonaM (approved):
Thanks for adding additional info!

PR #170, opened 2023-03-14
Wait of VMs to finish ACPI shutdown

justinc1:
Noticed that we do start ACPI shutdown. But VM with Porteus-XFCE-v5.0-x86_64.iso was still shutting down when we already send force shutdown. PR tries to fix this.


anazobec (approved):
Looks good, thank you.

PR #171, opened 2023-03-15
Ignore SSL protocol violation during certificate replacement

justinc1:
@ddemlow had problem replacing certificate, and this happened only it server  was access via VPN. Likely VPN resulted in just correct timing to raise unexpected SSL error - in python code we got "EOF occurred in violation of protocol (_ssl.c:997)".

PR catches ScaleComputingError exception, and checks if message had this particular message inside.
I didn't try to refactor the ScaleComputingError to have a dedicated exception - I'm not able to reproduce the problem, so it is hard to test any code modification. For same reason, it is hard to add any integration test for this.

The extra `module.warn` were added, to help debugging if anything similar happens later. I hope warn message is kind enough ("... ignore and continue") - I do not want to make anyone nervous.

### DIFF
--- a/plugins/modules/certificate.py
+++ b/plugins/modules/certificate.py
@@ -105,11 +105,39 @@ def ensure_present(
             TaskTag.wait_task(rest_client, task)
             break
         except ConnectionRefusedError:
+            module.warn(
+                f"retry {ii}/{max_retries}, ConnectionRefusedError - ignore and continue"
+            )
             sleep(2)
             continue
         except ConnectionResetError:
+            module.warn(
+                f"retry {ii}/{max_retries}, ConnectionResetError - ignore and continue"
+            )
             sleep(2)
             continue
+        except errors.ScaleComputingError as ex:
+            # Ignore "EOF occurred in violation of protocol (_ssl.c:997)"
+            # Also other have same problem - https://github.com/psf/requests/issues/3006#issuecomment-183394849.
+            # We do not use requests library, but problem is the same.
+            # ex.args - this is what Exception.__init__() stores into object.
+            ex_reason = ex.args[0]
+            # ex_reason is instance of SSLEOFError or ssl.SSLEOFError
+            strerror = ex_reason.strerror
+            if (
+                "EOF occurred in violation of protocol" in strerror
+                and "ssl.c" in strerror
+            ):
+                module.warn(
+                    f"retry {ii}/{max_retries}, SSLEOFError - ignore and continue"
+                )
+                sleep(2)
+                continue
+            else:
+                module.warn(
+                    f"retry {ii}/{max_retries}, re-raise unexpected exception" + str(ex)
+                )
+                raise
     after: TypedCertificateToAnsible = dict(certificate=get_certificate(module))
     return True, after, dict(before=before, after=after)
 
--- a/plugins/modules/certificate.py
+++ b/plugins/modules/certificate.py
@@ -116,28 +116,14 @@ def ensure_present(
             )
             sleep(2)
             continue
-        except errors.ScaleComputingError as ex:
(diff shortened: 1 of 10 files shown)


plugins/modules/certificate.py
```diff
@@ -105,11 +105,39 @@ def ensure_present(
...
+            )
             sleep(2)
             continue
         except ConnectionResetError:
+            module.warn(
+                f"retry {ii}/{max_retries}, ConnectionResetError - ignore and continue"
+            )
             sleep(2)
             continue
+        except errors.ScaleComputingError as ex:
```
domendobnikar, line 119: Can we add this to client.py in order to avoid string-to-string comparison?
I think the exception is called `SSLEOFError`.

domendobnikar (changes requested):

plugins/modules/certificate.py
domendobnikar: `            elif (
                e.args
                and isinstance(e.args, tuple)
                and type(e.args[0]) == SSLEOFError
            ):
                raise SSLEOFError(e)`

justinc1: Yes, it is called that way. But I have no way to test that code will still work after change. Replacing cert in loop, and in parallel accessing /rest/v1/ping shows `ConnectionRefused` and `ConnectionReset`, but I never got SSL error.

And this is problem - we refactor client.py to intercept/reraise one extra exception, and side effect might be this fix will no longer work.

domendobnikar: I'm wondering if we should just remove the raise Scale error part of the client.
We've had this issue a couple of times now, where we don't exactly know which exception we are trying to handle.
And that part of the code does not seem to add much value ...

For the testing part, would it be too much effort on our part to build and kill an SSL connection?
Maybe we don't even need to setup a server and just close the SSL connection to the actual cluster as the data is being sent.

justinc1: > just remove the raise Scale error 
Maybe. We would need to review all other `except`.

> to build and kill an SSL connection
Worth to try. But I expect a blind tcpkill will most likely break opened TCP connection before or after SSL handshake, before or after full record (record as in https://msatechnosoft.in/blog/secure-socket-layer-ssl-work-accomplish-task-ssl-services/). And that SSL will not complain - it already got all bytes of previous record, and is waiting on 1st byte of next record.

justinc1: How to reproduce that `EOF occurred in violation of protocol (_ssl.c:997)>`:
```
---
# ansi.yml 
- name: test SSL EOF
  hosts: all
  connection: local
  gather_facts: false

  tasks:
    - uri:
        url: https://127.0.0.1:44331
        validate_certs: false
```

```
# terminal 1
nc -l 127.0.0.1 44331

# terminal 2
ansible-playbook -i localhost, ansi.yml -v

# terminal 1
Ctrl+D or Ctrl+C

# terminal 2
TASK [uri] **********************************************************************************************************
fatal: [localhost]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python3"}, "changed": false, "elapsed": 2, "msg": "Status code was -1 and not [200]: Request failed: <urlopen error EOF occurred in violation of protocol (_ssl.c:997)>", "redirected": false, "status": -1, "url": "https://127.0.0.1:44331"}
```

No need to interrupt SSL handshake in middle. Just do not send any byte over wire.

domendobnikar: Nice work! 👍

justinc1:
Refactored exception handling. We expect about 3 different ssl errors.

Integ test passed in https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4479886132/jobs/7874449568

domendobnikar (approved):
Looks good, thank you. :+1:

PR #172, opened 2023-03-20
CI use SMB cleanup variables from integration_config.yml

justinc1:
Initially we had a configuration variable/secret for every possible line in `integration_config.yml`. Later we decided to have whole `integration_config.yml` except about 3 passwords as a single CI variable, and then we add only missing passwords to it.

The `smb_cleanup.sh` was not yet updated to get required configuration (SMB server, username etc) from `integration_config.yml`. This PR does this.

CI job to show that `smb_cleanup` integration "test" does pass - https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4465664492/jobs/7842970112

### DIFF
--- a/.github/workflows/integ-test.yml
+++ b/.github/workflows/integ-test.yml
@@ -239,4 +239,16 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: ${{ env.WORKDIR }}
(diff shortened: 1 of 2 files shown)


anazobec (approved):
Looks good, thank you. Approved.

PR #173, opened 2023-03-20
Add example for virtual_disk_info module

justinc1:
Because examples are nice to have. And I just needed to list all virtual_disks :)


domendobnikar (approved):
Looks good, thank you. 👍

PR #174, opened 2023-03-21
Handle case when update_status.json doesnt exist

PolonaM:
In case of “fresh cluster” no updates have ever been done so no update_status.json exists yet. Instead "404 Not found" is returned which caused module version_update_status_info (and consequently the version_update_single_node role) to fail.
With this PR the version_update_status_info module is fixed but the role version_update_single_node still needs to be revisited.


justinc1 (approved):
Thank you.

PR #176, opened 2023-03-22
CI Remove old half-uploaded virtual disk before upload

justinc1:
Half-uploaded virtual disk with name "uploading-ci-test-virtual-disk-1.qcow2" needs to be removed before we can upload a new "ci-test-virtual-disk-1.qcow2".


anazobec (approved):
Looks good. The CI docs tests are failing though, but the failure doesn't seem to be related to this minor change.

PR #177, opened 2023-03-22
CI fix ANSIBLE_COLLECTIONS_PATH for docs

justinc1:
Fixed typo

PR #178, opened 2023-03-22
Http put with content length

justinc1:
It is safer to always PUT binary data to HC3 with  Content-Length HTTP header.
PUT is used only in api, iso (and in future in virtual_disk module).

This does fix uploading virtual disk using raw api module example - `examples/virtual_disk.yml`.
It will also allow integration test for virtual_disk_info module to pass (the test VD is uploaded using api module).


justinc1:
#177 fixed the 'make docs'

PR #179, opened 2023-03-22
Add vm_snapshot_attach_disk module.

anazobec:
Add ``vm_snapshot_attach_disk`` module.

#### Current example usage:
```yaml
- name: Attach a disk from an automated (scheduled) VM snapshot to VM
  scale_computing.hypercore.vm_snapshot_attach_disk
    vm_name: example-vm
    source_snapshot_uuid: ec91ce38-a795-4c0b-bc72-60f8ddba8d91
    snapshot_type: automated  # "user" if the snapshot is not scheduled
    source_disk_slot: 0
```


plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+    type: str
+    description:
+      - Name of the VM we want to attach a VM snapshot disk to.
+    required: False
+  source_snapshot_uuid:
+    type: str
+    description:
+      - UUID of the snapshot we want to use on a VM.
+    required: False
+  snapshot_type:
```
justinc1, line 37: I would say `source_snapshot_type` - just to have `source_snapshot` everywhere.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+from hypercore.plugins.module_utils.task_tag import TaskTag
+
+DOCUMENTATION = r"""
+module: vm_snapshot_attach_disk
+
+author:
+  - Ana Zobec (@anazobec)
+short_description: Attach a disk from a snapshot to a VM on HyperCore API.
+description:
+  - Use this module to attach a disk from a snapshot to a desired VM on HyperCore API.
```
justinc1, line 20: Maybe add:
```
Specified disk from source snapshot will be cloned. Cloned dist is than attached to the destination VM.
```

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+      - Name of the VM we want to attach a VM snapshot disk to.
+    required: False
+  source_snapshot_uuid:
+    type: str
+    description:
+      - UUID of the snapshot we want to use on a VM.
+    required: False
+  snapshot_type:
+    type: str
+    choices: [ user, automated ]
```
justinc1, line 39: Is this lower or upper case? In vm_snapshot_info module - I thing it does return `USER`.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+    description:
+      - Specify a type that the chosen snapshot has to be.
+      - C(snapshot_type=user) - snapshot was created manually.
+      - C(snapshot_type=automated) - snapshot was created automatically through a schedule.
+    required: True
+  source_disk_slot:
+    type: int
+    description:
+      - Specify a snapshot disk slot to use on a VM.
```
justinc1, line 48: This sounds like it is slot on destination VM. But it is not. Change to:
```
      - Specify a disk slot from source snapshot to identify source disk.
```
Or to something better :)

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+    required: True
+"""
```
justinc1, line 50: We will need to also add parameters for destination VM. We already have `vm_name`, we need to add `vm_disk_type`, `vm_disk_slot`.

And one extra parameter for source - `source_disk_type`, as VM can have 1 SATA and 1 IDE disk, and then `source_disk_slot==0` can be either of two disks.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+  
+- name: Attach a disk from an user-made VM Snapshot to VM
+  scale_computing.hypercore.vm_snapshot_attach_disk
+    source_snapshot_uuid: ec91ce38-a795-4c0b-bc72-60f8ddba8d91
+    snapshot_type: user
+    vm_disk_slot: 0
+"""
+
+RETURN = r"""
+records:
```
justinc1, line 68: RETURN looks like old content from copy-paste.

What we can return? I guess VM - module changes destination VM. So return value will be same as for `vm` module.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+    module = AnsibleModule(
+        supports_check_mode=True,
+        argument_spec=dict(
+            arguments.get_spec("cluster_instance"),
+            vm_name=dict(
+                type="str",
+                required=True,
+            ),
+            source_snapshot_uuid=dict(type="str", required=True),
+            snapshot_type=dict(  # maybe we can throw out this one, since the user is already choosing a snapshot with its uuid and should already know if it's automated or not... I mean, the user knows what snapshot he's choosing
```
justinc1, line 184: You are right, this can go out.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+            chacheMode=source_disk_info["cache_mode"],
+            capacity=source_disk_info["capacity"],
+            shareUUID=source_disk_info["share_uuid"],
+            path=source_disk_info["path"],
+            slot=source_disk_info["slot"],
+            name=source_disk_info["name"],
+            disableSnapshotting=source_disk_info["disable_snapshotting"],
+            tieringPriorityFactor=source_disk_info["tiering_priority_factor"],
+        ),
+    )
```
justinc1, line 144: I think params `capacity`, `shareUUID`, `path`, `name` are read-only. Try to remove them from POST payload to see if you can omit them. Or check in UI what payload is sent to server for this action.

`disableSnapshotting` - I would try to omit that one too.

`type` - I think user should be able to select destination type. If I need to attach virtio disk, but my windows VM does not have virtio drivers, then type needs to be changed.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+
+    vm_snapshot = VMSnapshot.get_snapshots_by_query(
+        {"uuid": source_snapshot_uuid}, rest_client
+    )[0]  # there is only 1 snapshot with the specified uuid.
+    vm_uuid = vm_snapshot["vm"]["uuid"]
+
+    # TODO: add "device_snapshots" to module_utils/vm_snapshot.py
+    #  add "device_snapshots" dict to the vm_snapshot from_hypercore() method
+    #  add "vm.block_devices" dict to the vm_snapshot from_hypercore() method
+    source_disk_uuid = vm_snapshot["device_snapshots"][source_disk_slot]
```
justinc1, line 119: you will need to first filter by source_disk_type, then by source_disk_slot.

I think it is possible to have unused disk slots. Like used slots are 0,1,2, then delete disk 1, and used slots are 0,2.
Then simple '[ind]' will not work.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,209 @@
...
+    vm_name = module.params["vm_name"]  # maybe this parameter is not needed? We can get vm_uuid through the specified snapshot...
+    source_snapshot_uuid = module.params["source_snapshot_uuid"]
+    # snapshot_type = module.params["snapshot_type"]
+    source_disk_slot = module.params[
+        "source_disk_slot"
+    ]  # the higher the index, the newer the disk
+
+    vm_snapshot = VMSnapshot.get_snapshots_by_query(
+        {"uuid": source_snapshot_uuid}, rest_client
+    )[0]  # there is only 1 snapshot with the specified uuid.
```
justinc1, line 113: If uuid is invalid - make module exit with error message, not crash.

justinc1 (commented):
The direction is correct one.

If we ever need easier identification than `source_snapshot_uuid` - then we have open option to add extra `source_snapshot_xyz` parameters. So futre extension is possible, and that is nice.

plugins/modules/vm_snapshot_attach_disk.py
anazobec: Since this parameter is not really needed, I will throw it out.

plugins/module_utils/vm_snapshot.py
```diff
@@ -160,5 +180,38 @@ def filter_snapshots_by_params(
...
+        new_device_snaps = vm_snapshot.device_snapshots[:]
+
+        if filter["uuid"]:
+            new_device_snaps = [device_snap for device_snap in cls.device_snapshots if device_snap["uuid"] == filter["uuid"]]
+        if filter["slot"]:
+            new_device_snaps = [device_snap for device_snap in cls.device_snapshots if device_snap["slot"] == filter["slot"]]
+        if filter["type"]:
+            new_device_snaps = [device_snap for device_snap in cls.device_snapshots if device_snap["type"] == filter["type"]]
+
+        return new_device_snaps[0]
```
justinc1, line 217: What if there is more than 1 element in `new_device_snaps`? If you check for this case, and just raise ScaleComputingError (or subclass of it) - it will be safer.

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -0,0 +1,238 @@
...
+        dict(
+            before={}, after=created_block_device
+        ),  # before, we ofcourse, didn't have that new block device, and after we should have it
+    )
+
+
+def run(
+    module: AnsibleModule, rest_client: RestClient
+) -> List[Optional[TypedVMSnapshotToAnsible]]:
+    filtered = VMSnapshot.filter_snapshots_by_params(module.params, rest_client)
```
justinc1, line 195: A. I didn't notice until now `attach_disk` is not yet called. Hm...

justinc1 (commented):
I just tried to review changes.

plugins/modules/vm_snapshot_attach_disk.py
anazobec: It was indeed a copy-paste used as a placeholder for a now updated RETURN.

plugins/modules/vm_snapshot_attach_disk.py
anazobec: You are right. It is possible. I have used a proper filtering for this.

plugins/modules/vm_snapshot_attach_disk.py
anazobec: It should be called now, and it also should work :)
Unit tests and integration tests for this module (``vm_snapshot_attach_disk``) are on the way.

tests/integration/targets/vm_snapshot/tasks/helper_api_vm_snapshot_create.yml
```diff
@@ -1,15 +1,18 @@
...
+# ++++++++++++++++++++++++++++++
 - name: Create test VMs
   scale_computing.hypercore.vm:
     vm_name: "snapshot-test-vm-{{ item }}"
     description: Snapshot testing
     state: present
     tags:
       - Xlab_snapshot_testing
-    memory: "{{ '512 MB' | human_to_bytes }}"
+    memory: "{{ '4 GB' | human_to_bytes }}"
```
justinc1, line 12: Likely you do not need VM with more RAM for this test.

tests/unit/plugins/modules/test_vm_snapshot_info.py
justinc1, line 1: Thank you for adding unit test for _info module!

justinc1 (changes requested):
`vm_snapshot_info` returns in `records[0].block_devices` many fields, and some names/values are not compatible with what `vm_info` module does return. Should be fixed - for some fields, just do not return them.

Some disk type likely cannot be cloned (`nvram`, `vtpm`, etc). If it cannot be cloned, then do not offer option for cloning/attaching.

plugins/module_utils/vm_snapshot.py
```diff
@@ -158,10 +191,32 @@ def to_ansible(self) -> TypedVMSnapshotToAnsible:
...
+        check_vm = True  # it will be True if self.vm == None
+        if self.vm != {}:
+            vm_sorted_block_devices = [
+                dict(sorted(bd.items(), key=lambda item: item[0]))  # type: ignore
+                for bd in self.vm["block_devices"]
+            ]
+            other_sorted_block_devices = [
+                dict(sorted(bd.items(), key=lambda item: item[0]))  # type: ignore
+                for bd in other.vm["block_devices"]
+            ]
```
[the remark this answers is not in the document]
anazobec: Yes, I am trying to sort the ``block_devices`` dict by its key ``uuid``, not the whole ``vm``. With this, I ensure that when testing this class, I don't get a false negative because the ``block_devices`` keys are differently arranged. This is only needed for ``block_devices``.

plugins/module_utils/vm_snapshot.py
```diff
@@ -232,3 +307,83 @@ def send_delete_request(
...
+            if new_key == "uuid":
+                new_key = "block_device_uuid"
+            if new_key == "vir_domain_uuid":
+                new_key = "vm_uuid"
+
+            new_dict[new_key] = _dict[key]
+        return new_dict
+
+    @classmethod
+    def get_vm_disk_info_by_uuid(
```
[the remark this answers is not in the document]
anazobec: Hmmm... to get block device info only, I think it is easier to use ``rest/v1/virDomainBlockDevice/uuid`` otherwise another filter function would be needed to filter results returned by ``rest/v1/VirDomain/uuid``. For the last function in this file (``get_external_vm_uuid``) which does actually use the endpoint ``rest/v1/VirDomain/uuid`` I guess I could've just used the VM class instead... Well, as you've said, for now I will leave this code as is, and then later if needed, I will change it for optimization.

plugins/module_utils/vm_snapshot.py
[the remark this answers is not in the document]
anazobec: This was resolved. :)

justinc1 (changes requested):
Only one trivial typo to fix.

plugins/module_utils/vm_snapshot.py
```diff
@@ -158,10 +187,27 @@ def to_ansible(self) -> TypedVMSnapshotToAnsible:
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, VMSnapshot):
             return NotImplemented
+
+        check_vm = True  # it will stay True if self.vm == {}
+        if self.vm != {}:
+            vm_sorted_disks = sorted(self.vm["disks"], key=lambda disk: disk["iso_name"])  # type: ignore
+            other_sorted_disks = sorted(self.vm["disks"], key=lambda disk: disk["iso_name"])  # type: ignore
```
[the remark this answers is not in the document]
anazobec: Ah, I see now, thank you :D

justinc1 (approved):
ty

PR #186, opened 2023-03-29
Update return documentation - dict to list

PolonaM:
Updated return documentation type: dict changed to type: list


justinc1 (approved):
Thank you, I only changed commit message to include long story...

Issue #163, opened 2023-03-10
:rocket: Feature request: vm_clone module should provide a "preserve mac address" option like HyperCore UI


ddemlow:
## Is your feature request related to a problem? Please describe.
to simplify cloning of identical virtual machines from latest or roll back to a previous point in time snapshot - it would be helpful to have an option in vm_clone module that is the equivalent of the clone existing mac address option available in the web UI

## Describe the solution you'd like
simplify the process of cloning VM or from snapshot while retaining original mac addresses for virtual nics (which may be required to preserve dhcp reservations or to preserve static network configurations for some OS's)

## Describe alternatives you've considered
data could be extracted from vm_info, then clone task, then a vm_params task but this could be reduced down to just the clone task and be more declarative

## Additional context
test automation using virtual nodes using ansible could leverage this option

![image](https://user-images.githubusercontent.com/26821128/224410888-b3c5b48d-2ec9-4d58-a9ed-4063112f3362.png)

PR #187, opened 2023-03-29
Add preserve_mac_address option to vm_clone module

justinc1:
As requested in #163.
We want to be able to clone VM, and easily keep MAC address unmodified.

The class `NicType` was refactored, to be able to convert NIC type between ansible and HyperCore, in either direction. It does not need to be enum.Enum derived (in fact, I hade some problems with mypy - I think mypy was just not able to figure out what might methods return).

The `NIC._handle_nic_type` method is used in 2 places only. It will be replaced by `NicType.hypercore_to_ansible()`. The change is small, but it requires changing unit tests. Currently unittests mock data containt string "virtio" for HyperCore returned data - this is wrong, it should be "VIRTIO". The complete change is then bigger, so it will be in separate PR - #188.


anazobec (approved):
Looks good to me, thank you!

PR #188, opened 2023-03-30
Refactor Nic._handle_nic_type method

justinc1:
Draft: this requires #187 to be merged first - only last commit in refactor-NicType branch is relevant.

The `Nic._handle_nic_type` method implemented conversion of NIC type from hypercore to ansible.
But in code it was used also to "convert" from ansible to ansible - whis was wrong.
Unittest did pass because we had also wrong unittest data - mock data pretended HyperCore returns "virtio", but in reality "VIRTIO" is returned.

This is now changed, unittest mock data is fixed, and `Nic._handle_nic_type` is replaced with `NicType.hypercore_to_ansible`.


plugins/module_utils/type.py
```diff
@@ -8,11 +8,25 @@
...
+    v: k for k, v in FROM_HYPERCORE_TO_ANSIBLE_NIC_TYPE.items()
+}
 
 
 # Maybe create enums.py or scale_enums.py and move all enum classes there? @Jure @Justin
-class NicType(str, enum.Enum):
-    RTL8139 = "RTL8139"
-    VIRTIO = "VIRTIO"
-    INTEL_E1000 = "INTEL_E1000"
+class NicType:
```
PolonaM, line 25: I would just move this class to module_utils/nic.py (and delete type.py). But this is just my opinion, not necessary to change that. Otherwise nice job, I approve :)

PolonaM (approved):

PR #189, opened 2023-03-30
Add next and latest to return value

PolonaM:
To be able to choose next (the lowest number) or latest (the highest number) update from the list of available updates, next and latest are added to return value.

### DIFF
--- a/plugins/modules/version_update_info.py
+++ b/plugins/modules/version_update_info.py
@@ -22,36 +22,18 @@ extends_documentation_fragment:
 seealso:
   - module: scale_computing.hypercore.version_update
   - module: scale_computing.hypercore.version_update_status_info
-options:
-  select:
-    description:
-      - List desired version update from list of updates
-      - If C(next), then version with the lowest number in the list of available updates is returned
-      - If C(latest), then version with the highest number in the list of available updates is returned
-    type: str
-    choices: [ next, latest ]
 """
 
 EXAMPLES = r"""
 - name: Get a list of updates
   scale_computing.hypercore.version_update_info:
   register: result
-
-- name: Get a list of updates and next update
-  scale_computing.hypercore.version_update_info:
-    select: next
-  register: result
-
-- name: Get a list of updates and latest update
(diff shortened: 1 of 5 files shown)


plugins/modules/version_update_info.py
```diff
@@ -22,12 +22,30 @@
...
   - module: scale_computing.hypercore.version_update
   - module: scale_computing.hypercore.version_update_status_info
+options:
+  select:
+    description:
+      - List desired version update from list of updates
+      - If C(next), then version with the lowest number in the list of available updates is returned
+      - If C(latest), then version with the highest number in the list of available updates is returned
+    type: str
+    choices: [ next, latest ]
```
justinc1, line 32: I guess sometimes it is still nice to be able to list all available updated - as `all` choice. Main use case would be debugging, I guess.

Why not still returning the complete list, but sort it before it is returned? Then later in playbook you can `{{ (updates_result.records | last).uuid }}` or `{{ (updates_result.records | first).uuid }}`. I would sort them in ascending order (0,1,2,...) - is sort of natural.

plugins/modules/version_update_info.py
```diff
@@ -36,15 +54,77 @@
...
+      sample: 2
+    revision:
+      description: Revision number
+      type: int
+      sample: 11
+    timestamp:
+      description: Unix timestamp when the update was released
+      type: int
+      sample: 0
+record:
```
justinc1, line 90: Would be `records` list iff you accept the change

justinc1 (changes requested):
I think it is useful to be able to list all possible version updates.

plugins/modules/version_update_info.py
justinc1: Comment was invalid, `records` was unmodified, only `record` output was added.

justinc1 (approved):
Added trivial change to "Lowest update is "{{ result.next.uuid | default('') }}" to make sure example works also if no update is available.
I will squash merge the PR.

PR #192, opened 2023-03-31
Integ test

justinc1:
Just refactoring content in `tests/integration/integration-config.yml`. I want to be able to select SC_HOST for tests with a single line change. After this PR, we need to change only `sc_host`, the username and password are moved from top-level `.sc_username` to `.sc_config.<sc_host>.sc_username`.

In similar way, each `sc_host` will have a different replication destination - `sc_config.<sc_host>.sc_replication_dest_host`. The presence or absence of `sc_replication_dest_host` tells us if host was configured with replication or not. This will be used later to decide if replication related tests can be run.

In similar way we should be (later) able to add expected ICOS version, and expected next_update_version for each available test host. It is annoying to see tests are failing only because we cannot hard-code expected values.

Assigning everyone, as everyone needs to update local `integration-config.yml` too.


.github/actions/make-integ-config/action.yml
```diff
@@ -10,34 +10,39 @@ inputs:
     required: true
     # default: 'World'
   sc_username:
```
PolonaM, line 12: Change to sc_username_50 to make it more understandable

.github/actions/make-integ-config/action.yml
```diff
@@ -10,34 +10,39 @@ inputs:
...
-    description: 'sc_username variable'
+    description: 'sc_username variable to access physical nodes'
     required: true
   sc_password:
```
PolonaM, line 15: Change to sc_password_50 to make it more understandable

PolonaM (changes requested):

PR #193, opened 2023-03-31
Update return documentation - part2

PolonaM:


plugins/modules/vm_snapshot_info.py
```diff
@@ -77,17 +77,48 @@
...
-      type: USER
-      vm:
+  elements: dict
+  contains:
+    automated_trigger_timestamp:
+      description: Unix timestamp used when determining which automated snapshots to retain
+      type: int
+      sample: 0
+    block_count_diff_from_serial_number:
+      description: Snapshot serial number of the previous snapshot used to calculate VirDomainSnapshot.blockCountDiff # zakaj nimamo block_count_diff ?
```
justinc1, line 87: I will leave question, and add potential answer :). I think the meaning of `block_count_diff_from_serial_number` is `the previous snapshot`. Maybe that by itself has some use (I image create snapshot, restore from it, create a new snapshot, and then you end with non-sequential snapshot serial number.).

justinc1 (approved):
I will add trivial change or two, and squash-merge.

PR #197, opened 2023-04-07
Fix roles check_local_time, version_update_single_node

justinc1:
The job https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4625947983/jobs/8182193452#step:9:55 failed.
The `ansible.builtin.set_fact: executable="{{ executable_python }}"` got executed (I think), and then `executable.stdout_lines` - there is no `stdout_lines` in `executable`.

Whole 3 tasks are just to hard to understand, and also to test.

The PR will switch to use `discovered_interpreter_python`. I think this will just work, even if playbook would be executed over SSH.

The version_update_single_node assumed `version_update_status_info` always returns `record`. But if cluster was never upgraded, then no record is returned. This is fixed now.

The CI job for this example will still fail - we cannot really upgrade any of our test clusters.
And I do not have a good idea how to test that running the example on each test cluster failed in "correct" way.
Maybe I will just exclude that example from CI testing, or add a special CI job with more complicated shell commands.

### DIFF
--- a/examples/version_update_info.yml
+++ b/examples/version_update_info.yml
@@ -13,6 +13,6 @@
       ansible.builtin.debug:
         msg: >-
           Host has {{ result.records | length }} updates available.
-          All available updates: {{ result.records | map(attribute='uuid')}}.
+          All available updates: {{ result.records | map(attribute='uuid') }}.
           Lowest update is "{{ result.next.uuid | default('') }}".
           Highest update is "{{ result.latest.uuid | default('') }}".
--- a/examples/version_update_single_node.yml
+++ b/examples/version_update_single_node.yml
@@ -10,6 +10,10 @@
     desired_version: 9.1.23.210897
 
   tasks:
+    - name: Show desired_version
+      ansible.builtin.debug:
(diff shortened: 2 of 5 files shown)


PolonaM:
I replaced "discovered_interpreter_python" with "ansible_python_interpreter" because the first one was not working for me. I also discovered that now that we are in CEST, the integration test were failing. The reason was that ansible_date_time.tz returns CEST which is not a valid timezone for env var "TZ", and if TZ is an invalid timezone then time.localtime() will return UTC time. I Fixed this by getting timezone for the computer on which the test are running from https://ipapi.co/timezone which returns the timezone in shape of "Europe/Ljubljana" instead of CEST

justinc1:
`ansible_python_interpreter` should be better than `discovered_interpreter_python`. I'm not even sure where I got that variable... I would say - a very stupid typo from my side.

Using external resources in tests is in general bad practice - here to get timezone of current VM/container. Reason - sometimes you external site is not reachable, or is down, or might just die. But I will not complain to much. If needed, we can still refactor tests.

PR #199, opened 2023-04-11
deprecation for vm and snapshot_schedule return value

justinc1:
Running `ansible-playbook examples/vm.yml` shows 

```
TASK [Create VM demo-vm] ******************************************************************************
[DEPRECATION WARNING]: The 'record' return value will be changed from list (containing a single item) to dict. There will be no release where both old and 
new variant work at the same time. To ease migration, the only change between last 1.x or 2.x release and 3.0.0 release will be changing the 'record' 
return value. Affected modules are vm and snapshot_schedule. This feature will be removed from scale_computing.hypercore in version 3.0.0. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
ok: [localhost]
```

Examples for `vm` and `snapshot_schedule` modules were added.
Examples for `vm`, `snapshot_schedule`, `iso` and `vm_nic` contain example task for old and new syntax.

The same example tasks are show in `deprecation.rst`, a document that is support to help users migrate their playbooks.

Space for existing deprecation messages are fixed.

Relative links are used in `deprecation.rst`, so we can render documentation locally, and link will guide to locally rendered modules. Relative links were added to `quickstart.rst` too.

After merge to milestone-1-done branch (e.g. will be in release 1.2.0), I will cherry-pick this for main branch too.

### DIFF
--- a/plugins/modules/iso.py
+++ b/plugins/modules/iso.py
@@ -52,6 +52,8 @@ notes:
   - C(check_mode) is not supported.
   - Return value C(record) is added in version 1.2.0, and deprecates return value C(results).
     Return value C(results) will be removed in future release.
+    R(List of deprecation changes, scale_computing.hypercore.deprecation)
+    includes examples to help with transition.
 """
 
 
--- a/plugins/modules/iso.py
+++ b/plugins/modules/iso.py
@@ -256,8 +256,8 @@ def main():
     )
 
     module.deprecate(
-        "The 'results' return value is being renamed to 'record' and changed from list to dict."
-        "Please use 'record' since 'results' will be removed in future release."
+        "The 'results' return value is being renamed to 'record' and changed from list to dict. "
+        "Please use 'record' since 'results' will be removed in future release. "
         "But for now both values are being returned to allow users to migrate their automation.",
         version="3.0.0",
         collection_name="scale_computing.hypercore",
--- a/docs/rst/extra/deprecation.rst
+++ b/docs/rst/extra/deprecation.rst
@@ -1,0 +1,169 @@
+.. _scale_computing.hypercore.deprecation:
+
+*****************
+Deprecation Notes
+*****************
+
+Deprecation notes will list deprecated features in different HyperCore Collection releases
+and guide users through migration.
+
+Release 1.0.0
+=============
+
+Initial release, no deprecated features.
+
+Release 1.1.0
+=============
+
+No deprecated features.
+
+Release 1.2.0
+=============
+
+Module `scale_computing.hypercore.iso <../collections/scale_computing/hypercore/iso_module.html>`_
+return value ``results`` is deprecated.
+A new return value ``record`` is added as replacement.
+The old ``results`` return value will remain available until release 3.0.0.
+
+Module `scale_computing.hypercore.vm_nic <../collections/scale_computing/hypercore/vm_nic_module.html>`_
+return value ``records`` is deprecated.
+A new return value ``record`` is added as replacement.
+The old ``records`` return value will remain available until release 3.0.0.
+
+Only playbooks that store output from ``scale_computing.hypercore.iso``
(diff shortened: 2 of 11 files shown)


plugins/modules/iso.py
```diff
@@ -256,8 +256,8 @@ def main():
     )
 
     module.deprecate(
-        "The 'results' return value is being renamed to 'record' and changed from list to dict."
-        "Please use 'record' since 'results' will be removed in future release."
+        "The 'results' return value is being renamed to 'record' and changed from list to dict. "
+        "Please use 'record' since 'results' will be removed in future release. "
         "But for now both values are being returned to allow users to migrate their automation.",
```
PolonaM, line 261: Should we also have "affected module: iso" in the deprecation message (similar to vm)? The same goes for vm_nic

PolonaM (changes requested):

plugins/modules/iso.py
justinc1: Could be a bit confusing, for `iso` and `vm_nic` we are able to implement new way before old way of working is removed. What if I instead replace in vm/snapshot_schedule module (and add to iso/vm_nic) :

```
# remove
# DOCUMENTATION, notes section
    Affected modules are M(scale_computing.hypercore.vm) and M(scale_computing.hypercore.snapshot_schedule).
#    module.deprecate later
        "Affected modules are vm and snapshot_schedule.",

# add
# DOCUMENTATION, notes section
        "For complete list of deprecation changes, please see https://justinc1.github.io/HyperCoreAnsibleCollection/extra/deprecation.html",
#    module.deprecate later
        "For complete list of deprecation changes, please see https://justinc1.github.io/HyperCoreAnsibleCollection/extra/deprecation.html",
```

We add one `deprecation.rst` file, and there we can include details, also an example how new playbook using `vm` module will look like.

justinc1: CC @jtaubenheim - I expect you also have opinion about this.

jtaubenheim: @justinc1 as I said in the slack convo I do not believe we should remove the old instructions on the module pages until 3.0.0 release (which you have noted). My only question is if there will be some note or warning on the module documentation pages (i.e., https://scalecomputing.github.io/HyperCoreAnsibleCollection-docs/modules/snapshot_schedule.html) <--- this is still the old version on AnsibleHyperCoreCollection-docs. But at note that directs users from the affected modules details page to the deprecation page? I don't see any updates to the files `/snapshot_schedule.rst.txt` or `/vm.rst.txt`. My only concern here is that we cannot expect users to regularly check the deprecation page and it needs to be noted on the affected module pages that they need to visit the deprecation page to see the details.

justinc1: @jtaubenheim I didn't correctly understand that you are asking to include link to deprecation page on affected module pages.

It actually is already there, in "notest" section of each module page. I will just paste all links (for my fork):
- https://justinc1.github.io/HyperCoreAnsibleCollection/collections/scale_computing/hypercore/iso_module.html#notes
- https://justinc1.github.io/HyperCoreAnsibleCollection/collections/scale_computing/hypercore/vm_nic_module.html#notes
- https://justinc1.github.io/HyperCoreAnsibleCollection/collections/scale_computing/hypercore/vm_module.html#notes
- https://justinc1.github.io/HyperCoreAnsibleCollection/collections/scale_computing/hypercore/snapshot_schedule_module.html#notes

Common text to all is "[List of deprecation changes](https://justinc1.github.io/HyperCoreAnsibleCollection/extra/deprecation.html#scale-computing-hypercore-deprecation) includes examples to help with transition." Will this be sufficient?

jtaubenheim: Yep that looks good. I just didn't want to rely on people navigating to the deprecation page to know that there are planned changes to the specific modules and having it on the module pages is really just a CYA precaution. Looks good to me!

justinc1: Thank you, merging now.

PR #203, opened 2023-04-12
Fix version update role

justinc1:
Two corner cases fixed.

1st if cluster was never upgrade before, the `version_update_status_info` module will return `record=None`. 

2nd, after one upgrade failed (upgrade was initiated, after a while is stopped with "Failed to open update tunnel") the `version_update_status_info` module does return  `record.update_status=None`.

### DIFF
--- a/tests/integration/targets/version_update/tasks/main.yml
+++ b/tests/integration/targets/version_update/tasks/main.yml
@@ -24,9 +24,13 @@
     - name: Get status of the latest update applied
       scale_computing.hypercore.version_update_status_info:
       register: update_status
+    # Fail early if update is in progress
     - ansible.builtin.assert:
         that:
-          - update_status.record.update_status == "COMPLETE" or update_status.record.update_status == "TERMINATING" # Fail early if update in progress
+          - >-
+            update_status.record == None or
+            update_status.record.update_status == "COMPLETE" or
+            update_status.record.update_status == "TERMINATING"
 
     - name: Update to the same hc version
       scale_computing.hypercore.version_update:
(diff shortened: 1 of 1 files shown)


tests/integration/targets/version_update/tasks/main.yml
```diff
@@ -24,9 +24,25 @@
...
+    #        }
+    #    }
+    #}
     - ansible.builtin.assert:
         that:
-          - update_status.record.update_status == "COMPLETE" or update_status.record.update_status == "TERMINATING" # Fail early if update in progress
+          - >-
+            update_status.record == None or
+            update_status.record.update_status == None or
+            update_status.record.update_status == "COMPLETED" or
```
PolonaM, line 44: Here should be COMPLETE (but I agree, "completed" sounds better :))

tests/integration/targets/version_update/tasks/main.yml
```diff
@@ -24,9 +24,25 @@
...
+    #            "usernotes": "Failed to open update tunnel"
+    #        }
+    #    }
+    #}
     - ansible.builtin.assert:
         that:
-          - update_status.record.update_status == "COMPLETE" or update_status.record.update_status == "TERMINATING" # Fail early if update in progress
+          - >-
+            update_status.record == None or
+            update_status.record.update_status == None or
```
PolonaM, line 43: Nice catch!

PolonaM (changes requested):

tests/integration/targets/version_update/tasks/main.yml
justinc1: A, one more typo. Indeed, I see this `update_status.json` content:
```
{
    "prepareStatus": "",
    "updateStatus": {
        "masterStateID": "4",
        "masterState": "COMPLETE",
        "fromBuild": "208456",
        "toBuild": "211326",
        "toVersion": "9.1.24.211326",
        "currentComponent": "728",
        "totalComponents": "728",
        "percent": "100",
        "status": {
            "component": "7690\/9999",
            "node": "172.31.7.11",
            "statusdetails": "Update Complete. Press 'Reload' to reconnect",
            "usernotes": "Update Complete. Press 'Reload' to reconnect"
        }
    }
}
```

Testing this is just to hard...

PR #206, opened 2023-04-13
Do not deprecate vm_nic records return value

justinc1:
The module does work with list of NICs (the items input parameter). It is natural to also return a list of NICs.

Reverts db9dff797b6f01fe624ca1227139a836fda04e50

Draft: until at least one more person agrees for vm_nic module `records` is ok, and `record` is not.


plugins/modules/vm_nic.py
```diff
@@ -140,37 +136,6 @@
...
-    connected:
-      description: Enabled and can make connections
-      type: bool
-      sample: true
-    ipv4_addresses:
-      description: IPv4 addresses registered with this device
-      type: list
-      elements: str
-      sample: 192.0.2.1
 records:
```
PolonaM, line 139: Here the "type" needs to be list? Also the second line under "description" is not needed anymore

tests/unit/plugins/modules/test_vm_nic.py
```diff
@@ -200,7 +200,6 @@ def test_minimal_set_of_params(self, run_main_with_reboot, mocker):
         assert results == {
             "changed": False,
             "records": {},
```
PolonaM, line 202: Should there be an empty list? In the case of ensure_absent() an empty list is returned

PolonaM (changes requested):

plugins/modules/vm_nic.py
justinc1: List is ok, a single NIC can have multiple IP addresses.

PolonaM: Hm, I can see now that it is not clear that this question is meant for "records"

tests/unit/plugins/modules/test_vm_nic.py
justinc1: I guess we have incorrectly mocked return values. I do not want to extend this PR with fixing unittests.

plugins/modules/vm_nic.py
justinc1: Removed the second line from "description", thank you for noticing.

PolonaM (approved):

PR #210, opened 2023-04-14
Cluster shutdown module exception handling update

### LINKED ISSUE #137: Raise exception within module
Check every module for exception handling.
We should not be checking string error messages => in case formatting is different it wont work.
cluster_shutdown; SMTP; oidc_config

PolonaM:
Use ConnectionRefusedError in cluster shutdown instead of error description comparison

Closes #137

### DIFF
--- a/plugins/module_utils/cluster.py
+++ b/plugins/module_utils/cluster.py
@@ -11,7 +11,7 @@ __metaclass__ = type
 from time import sleep
 from ..module_utils.utils import PayloadMapper
 from ..module_utils.rest_client import RestClient
-from ..module_utils.errors import ScaleComputingError
+from ..module_utils.errors import ScaleComputingError, ScaleTimeoutError
 from ..module_utils.typed_classes import TypedClusterToAnsible, TypedTaskTag
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


plugins/module_utils/cluster.py
```diff
@@ -84,6 +84,9 @@ def shutdown(
                 dict(forceShutdown=force_shutdown),
                 check_mode,
             )
+        # TimeoutError is handled in rest_client.create_record:
+        # except TimeoutError as e:
+        #       raise errors.ScaleComputingError(f"Request timed out: {e}")
         except ScaleComputingError as e:
             # To avoid timeout when there are a lot of VMs to shutdown
             if "Request timed out" in str(e):
```
justinc1, line 92: Here we also detect exception root cause from string, not from type. Are the more places with this pattern?

We should have a dedicated exception, like ScaleRequestTimedOut. Then you refactor the RestClient, and you need to check if there is any other module that needs to know about your new exception.

justinc1 (changes requested):
Change is ok, but more such changes are needed.

plugins/module_utils/cluster.py
PolonaM: But this error will never cause a problem right? The string is generated inside the client, so the string comparison will never fail due to formatting (like it would when we compared the string directly from the http response - in case of "connection refused")

PolonaM: As discussed, I implemented ScaleTimeoutError in rest_client

justinc1 (approved):
ok and thank you.

Issue #198, opened 2023-04-07
:rocket: Feature request: shutdown vm's with specific tags before cluster update


ddemlow:
## Is your feature request related to a problem? Please describe.
this is somewhat a corner case for now (in some cases testing potential future capabilities) - but I have clusters with specific VM's that can't live migrate to other nodes in the cluster causing cluster rolling updates to fail if they are running.  examples include vm's running certain nested virtualization stacks (like nested hypercore nodes), with device passthrough like gpu or usb or could even be if strict affinity is set only allowing vm to run on specific node either due to licensing restrictions or where failover is not required or desired - for example a kubernetes node in multi-node kubernetes cluster

## Describe the solution you'd like
some way to automate the shutdown of those VM's before rest of cluster update is initiated

## Describe alternatives you've considered
add a task or role (either independent or into existing update roles) that loops through running VM's matching specified tag(or possibly tags?) - goes through shutdown process (initiate apci, wait timeout,force off if needed) - then continues

## Additional context
really just looking for help / best practice for including this in existing update roles / examples or creating new example

PR #218, opened 2023-04-25
Shutdown tagged vms

justinc1:
We already had a role with taskfile `shutdown_vms` to shutdown all VMs. This is used to shutdown all VMs in a single node cluster before version upgrade, and to start them back after upgrade.

The PR extended `shutdown_vms` with option to shutdown only some VMs. VM tag is used to decide if VM needs to be shutdown. This will be used during multinode cluster version upgrade to shutdown VMs than should not be live-migrated to another node. See also #198.

Integration test in https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4799944517


anazobec (approved):
Looks good, thank you.

PR #220, opened 2023-04-26
Fix vm_clone to clone replicated VM

justinc1:
The VirDomain/<uuid>/clone payload must include snapUUID. This required extengin VM class with snapshot_uuids field (and extending many existing unit tests).

Integration test for cloning replicated VM on destination cluster was added. This required adding sc_replication_dest_cluster_name to integration_config.yml.

I noticed vm_snapshot_info could include into output also `unavailable` attribute. VM snapshot cannot be used for clonning until it gets available. This could/should be done in another PR.


domendobnikar:
Integration tests:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/4912332605/jobs/8771240402

domendobnikar (approved):
Looks good, thank you. 👍

PR #223, opened 2023-05-03
Adding vm_snapshot examples to collection:

domendobnikar:
- filter and delete based on special properties
- filter and delete based on timestamps

### DIFF
--- a/examples/vm_snapshot_removal_based_on_timestamp.yml
+++ b/examples/vm_snapshot_removal_based_on_timestamp.yml
@@ -1,0 +1,68 @@
+---
+- name: Example - delete all snapshots that are older than X days.
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  environment:
+    SC_HOST: "your_host"
+    SC_USERNAME: "your_username"
+    SC_PASSWORD: "your_password"
+  vars:
+    # Format: 'YYYY-MM-DD hh:mm:ss'
+    # All snapshots older than this date will be deleted.
+    use_date: '2023-05-03 12:52:00'
+
+  tasks:
+  # ------------------------------------------------------
+    - name: List all snapshots
+      scale_computing.hypercore.vm_snapshot_info:
+      register: snapshot_results
+
+    - name: Convert date to unix timestamp "epoch"
+      set_fact:
+        epoch_timestamp: "{{ (use_date | to_datetime).strftime('%s') }}"
+
+    - name: Show epoch_timestamp
+      ansible.builtin.debug:
+        var: epoch_timestamp
+
+    - name: Create filtered_snapshots list
+      set_fact:
+        filtered_snapshots: []
+
+    - name: Loop through snapshots and add snapshots that are older than "use_date"
+      set_fact:
+        filtered_snapshots: "{{ filtered_snapshots + [ item ] }}"
+      when: "{{ item.timestamp  > epoch_timestamp | int }}"
+      loop: "{{ snapshot_results.records }}"
+      no_log: true
+
+    - name: Show only snapshots that are older than "use_date"
+      ansible.builtin.debug:
+        var: filtered_snapshots
+
+    # We could reuse "filtered_snapshots" here instead of "snapshot_results" and avoid the "when" statement.
+    # But leaving it as is for example purposes.
+    # Since this is the only mandatory task of the playbook, can be copy-pasted and reused as standalone task.
+    - name: Loop through list of snapshots and delete all older than the "use_date"
+      scale_computing.hypercore.vm_snapshot:
+        vm_name: "{{ item.vm.name }}"
+        label: "{{ item.label }}"
+        state: absent
+      when: "{{ item.timestamp > epoch_timestamp | int }}"
+      loop: "{{ snapshot_results.records }}"
+
+    - name: Create filtered_snapshots list - second time
+      set_fact:
+        filtered_snapshots: []
+
+    - name: Loop through snapshots and add snapshots that are older than "use_date" - second time
+      set_fact:
+        filtered_snapshots: "{{ filtered_snapshots + [ item ] }}"
+      when: "{{ item.timestamp  > epoch_timestamp | int }}"
+      loop: "{{ snapshot_results.records }}"
+      no_log: true
+
+    - name: Show only snapshots that are older than "use_date" - second time
+      ansible.builtin.debug:
+        var: filtered_snapshots
--- a/examples/vm_snapshot_special_cleanup.yml
+++ b/examples/vm_snapshot_special_cleanup.yml
@@ -1,0 +1,60 @@
+---
+- name: Example - delete all snapshots with label "TEST" and type "USER".
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  environment:
+    SC_HOST: "your_host"
+    SC_USERNAME: "your_username"
+    SC_PASSWORD: "your_password"
+
+  tasks:
+  # ------------------------------------------------------
+    - name: List all snapshots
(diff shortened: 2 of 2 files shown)


domendobnikar:
vm_snapshot PR needs to be merged first.
Sanity issues are related to missing module.

examples/vm_snapshot_removal_based_on_timestamp.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  tasks:
+  # ------------------------------------------------------
+    - name: List all snapshots
+      scale_computing.hypercore.vm_snapshot_info:
+      register: snapshot_results
+
+    - name: Convert date to unix timestamp 'epoch'
+      set_fact:
+        epoch_timestamp: "{{ (use_date | to_datetime).strftime('%s') }}"
```
justinc1, line 23: Timezone could be problematic here.

I tried to add timezone to date. Test task:
```
    - name: Show date
      ansible.builtin.debug:
        msg: |
          {{ ('2023-05-03 12:52:00' | to_datetime).strftime('%s') }}
          {{ ('2023-05-03 12:52:00+0000' | to_datetime('%Y-%m-%d %H:%M:%S%z')).strftime('%s') }}
          {{ ('2023-05-03 12:52:00+0200' | to_datetime('%Y-%m-%d %H:%M:%S%z')).strftime('%s') }}
          {{ ('2023-05-03 12:52:00 UTC' | to_datetime('%Y-%m-%d %H:%M:%S %Z')).strftime('%s') }}
          {{ ('2023-05-03 12:52:00 CEST' | to_datetime('%Y-%m-%d %H:%M:%S %Z')).strftime('%s') }}
          {{ ('2023-05-03 12:52:00 CET' | to_datetime('%Y-%m-%d %H:%M:%S %Z')).strftime('%s') }}
#          {{ ('2023-05-03 12:52:00 EST' | to_datetime('%Y-%m-%d %H:%M:%S %Z')).strftime('%s') }}
#          {{ ('2023-05-03 12:52:00 EDT' | to_datetime('%Y-%m-%d %H:%M:%S %Z')).strftime('%s') }}
#          {{ ('2023-05-03 12:52:00 America/Los_Angeles' | to_datetime('%Y-%m-%d %H:%M:%S %Z')).strftime('%s') }}
```

Last 3 lines generated error for me (ansible 2.14.5, python 3.10). Output from other lines was not what I would expect:
```
TASK [Show date] ****************************************************
ok: [localhost] => 
  msg: |-
    1683111120
    1683111120
    1683111120
    1683111120
    1683111120
    1683111120
```

Looks like timezone info is accepted, and then just ignored :/.

Can we document that `use_date` should be specified in UTC timezone? Is it UTC?

If needed, we can later use some code from check_local_time role, and do our own timezone aware conversion.

examples/vm_snapshot_special_cleanup.yml
```diff
@@ -0,0 +1,60 @@
+---
+- name: Example - delete all snapshots with label "TEST" and type "USER".
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  environment:
+    SC_HOST: "your_host"
+    SC_USERNAME: "your_username"
+    SC_PASSWORD: "your_password"
```
justinc1, line 9: Not needed.

examples/vm_snapshot_removal_based_on_timestamp.yml
```diff
@@ -0,0 +1,68 @@
...
+  connection: local
+  gather_facts: false
+  environment:
+    SC_HOST: "your_host"
+    SC_USERNAME: "your_username"
+    SC_PASSWORD: "your_password"
+  vars:
+    # Format: 'YYYY-MM-DD hh:mm:ss'
+    # All snapshots older than this date will be deleted.
+    use_date: '2023-05-03 12:52:00'
```
justinc1, line 13: Change default `use_date` to some "safer" value. If I run example unintentionally, I will delete snapshots, and that can be a problem. Something like `1999-05-03 12:52:00`?

examples/vm_snapshot_special_cleanup.yml
```diff
@@ -0,0 +1,60 @@
...
+      register: snapshot_results
+
+    - name: Create filtered_snapshots list
+      set_fact:
+        filtered_snapshots: []
+
+    - name: Loop through snapshots and add snapshots with label 'TEST' and type 'USER' to filtered_snapshots
+      set_fact:
+        filtered_snapshots: "{{ filtered_snapshots + [ item ] }}"
+      when: "{{ item.label  == 'TEST' and item.type == 'USER' }}"
```
justinc1, line 24: Make `'TEST'` a variable with name like `snapshot_label`. Then example can actually be used.

justinc1 (changes requested):
Minor changes.

The timezone is problematic, but if it is always UTC, than we just document this.

examples/vm_snapshot_removal_based_on_timestamp.yml
domendobnikar: Well, the main thing is that the time specified in the ansible playbook has to match the timezone set on the Scale cluster.
As long as that is the case there shouldn't be any timezone limitations set on our side.
We can add a comment that says something like "use_date timezone should match Scale cluster timezone".

domendobnikar:
We talked about adding a practical example of using uuid with vm_snapshot. @justinc1 
I added examples with absent (using uuid) into both of these example playbooks.
Would something like this be enough? Or should I add a separate playbook that shows both preset and absent?

plugins/modules/vm_snapshot.py
```diff
@@ -145,6 +150,31 @@
 from ..module_utils.vm import VM
 
 
+def get_snapshot(
```
justinc1, line 153: Any chance to get a snapshot with correct label, but from a different VM?

It seems to me: filter to `domainUUID=vm_object.uuid` part is lost. And integration tests should be extended too.

justinc1 (commented):
I'm worried `get_snapshot()` has small bug.

plugins/modules/vm_snapshot.py
domendobnikar: We pass VMobject as one of the parameters in this function.
It is then used with label for snapshot query if snapshot_uuid wasn't passed as parameter to Ansible.
 The only thing I added to this part was the `if module.params["uuid"]:` , everything else is the same.
 UUID is unique throughout the platform and use a different endpoint.

domendobnikar:
Integration tests:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5003882224

justinc1 (commented):
Added integration test - OK, thank you.

I think I didn't manage to describe what was potential problem. Say we have:
- vm_a, with snapshot:
  - snap_label_X, snap_uuid_a
- vm_b, with snapshot:
  - snap_label_X, snap_uuid_b

Then I try to task:
```
vm_snaptshot:
  vm_name: vm_a
  snapshot_uuid: snap_uuid_b
```
I think this case is not covered by current integration tests. In this case, module might continue to work with snap_uuid_b, but it should fail, vm_a has no snap_uuid_b.

domendobnikar:
Oh okey, I misunderstood the problem.
I will extend tests to also cover that potential bug.

domendobnikar:
Integration tests:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5013834038

domendobnikar:
Integration tests:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5014079719

justinc1 (approved):
ty

PR #224, opened 2023-05-08
Fix update status check

justinc1:
We had a race condition. Sometimes a magic 60 sec delay was not enough to get update/update_status.json API endpoint work. The version_update_status_info now (it didn't initially) returns None in this case, so we can use that.


anazobec (approved):
Looks fixed. Thank you.

PR #226, opened 2023-05-08
Fix version_update_single_node to send shutdown request multiple times.

anazobec:
Some VMs didn't want to receive a shutdown request, so ``version_update_single_node`` was updated so that the shutdown request is sent multiple times - in a loop.

### DIFF
--- a/roles/version_update_single_node/tasks/main.yml
+++ b/roles/version_update_single_node/tasks/main.yml
@@ -57,6 +57,7 @@
       include_tasks: shutdown_vms.yml
       vars:
         scale_computing_hypercore_shutdown_vms: "{{ vm_info }}"
+      loop: "{{ range(0, (scale_computing_hypercore_shutdown_wait_time / 10.0) | round(0, 'ceil') | int) | list }}"
       when: vms.records != []
 
     # ----------------- UPDATE --------------------
--- a/roles/version_update_single_node/tasks/main.yml
+++ b/roles/version_update_single_node/tasks/main.yml
@@ -57,8 +57,7 @@
       include_tasks: shutdown_vms.yml
       vars:
         scale_computing_hypercore_shutdown_vms: "{{ vm_info }}"
-      loop: "{{ range(0, (scale_computing_hypercore_shutdown_wait_time / 10.0) | round(0, 'ceil') | int) | list }}"
-      when: vms.records != []
+      when: scale_computing_hypercore_shutdown_vms.records != []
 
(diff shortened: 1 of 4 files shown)


justinc1:
For some reason Windows VMs sometimes ignore ACPI shutdown. Current PR would:
- rapidly send multiple shutdown request
- wait 300 sec on VMs to shutdown

It would be better if send shutdown is moved into wait on VM shutdown loop. So if some VM does not shutdown, then we retry shutdown. That is inside `shutdown_vms.yml`, move taks `name: Shutdown running VMs` into:

```
# Wait up to 300 sec (30*10)
- name: Wait until VMs shutdown
  include_tasks: wait_vm_shutdown.yml
  loop: "{{ range(0, 30) | list }}"
  when: version_update_all_vms_stopped | default(true)
``` 

Would that work?

tests/integration/targets/role_version_update_single_node/tasks/main.yml
```diff
@@ -1,8 +1,8 @@
 ---
 - environment:
     SC_HOST: "{{ sc_host }}"
-    SC_USERNAME: "{{ sc_config[sc_host].sc_username }}"
-    SC_PASSWORD: "{{ sc_config[sc_host].sc_password }}"
+    SC_USERNAME: "{{ sc_username_50 }}"
+    SC_PASSWORD: "{{ sc_password_50 }}"
```
justinc1, line 5: I will revert that change, it would prevent testing role on different clusters. For CI tests there is idea to select with  `sc_host` the test cluster, then all other relevant data/configuration needed for CI test are in `integration_config.yml`, at `sc_config[sc_host].xyz`.

I believe you wanted to add a whole additional host to your `integration_config.yml` (one of out test NUCs). You can do that, we intentionally commit to git only `integration_config.yml.j2`, so we can have extra hosts in `integration_config.yml`.

roles/version_update_single_node/tasks/main.yml
```diff
@@ -57,7 +57,7 @@
       include_tasks: shutdown_vms.yml
       vars:
         scale_computing_hypercore_shutdown_vms: "{{ vm_info }}"
-      when: vms.records != []
+      when: scale_computing_hypercore_shutdown_vms.records != []
```
justinc1, line 60: Thank you for fixing my bug :)

justinc1 (approved):
Thank you. I will merge, and later test manually.

PR #227, opened 2023-05-09
Update release workflow

justinc1:
Release workflow is updated with 2 extra steps:
- top-level README.md needs to be updated
- documentation needs to be updated

The galaxy.yml in main branch shouldn't contain already released version. Following some other examples, we can have there a to-be-released version with extra "-dev0" suffix.


domendobnikar (approved):
Looks good, thank you 👍

PR #232, opened 2023-05-15
Fix cloud-init meta-data/user-data content being mixed

justinc1:
A number of examples were including user-data content into meta-data section, and user-data section containing a bogus meta-data content.

Change this, examples should be close to valid and runnuble example.

Also add extra note to user-data example file that the file must begin with magic comment. This is always very easy to miss.


justinc1:
@PolonaM recently used cloud-init, and having "a bit wrong" examples was confusing. The PR will hopefully avoid this confusion for end-users.

@jtaubenheim is a reviewer too, since he authored quickstart guide, and PR does (trivial) modification to it.

PR #233, opened 2023-05-15
Add shutdown destination VM functionallity before attaching a disk from a source snapshot.

anazobec:
This PR makes sure that the VM we are trying to attach a snapshot disk to shuts down before attaching; tasks fail if we are trying to attach to a running VM.

**How it works (steps)**:
1. if the destination VM is running, shut it down (first normally, and if that fails, then forcibly) - this step is skipped if the VM is already shut down.
2. attach the desired snapshot disk
3. restart the destination VM.

### DIFF
--- a/plugins/modules/vm_snapshot_attach_disk.py
+++ b/plugins/modules/vm_snapshot_attach_disk.py
@@ -107,10 +107,14 @@ from ..module_utils import errors, arguments
 from ..module_utils.client import Client
 from ..module_utils.rest_client import RestClient
 from ..module_utils.vm_snapshot import VMSnapshot
+from ..module_utils.vm import VM
 from ..module_utils.task_tag import TaskTag
 from ..module_utils.typed_classes import TypedDiff
 from typing import Tuple, Dict, Any, Optional
 
+
+# A minimum size of an IDE typed disk in HC3 API is 1 GB
+MIN_TYPE_IDE_SIZE = 1000341504  # bytes
 
 # ++++++++++++
 # Must be reviewed - not sure if that's how this should work
@@ -164,9 +168,22 @@ def attach_disk(
             dict(before=before_disk, after=None),
         )
 
+    # Get VM before attach
+    # vm = VM.get_by_name(module.params, rest_client, must_exist=True)  # type: ignore
+
+    # First power off the destination VM
+    VMSnapshot.power_off_vm(module, rest_client)
+
     source_disk_info = VMSnapshot.get_snapshot_disk(
         vm_snapshot, slot=source_disk_slot, _type=source_disk_type
     )
+
+    destination_size = source_disk_info["size"]
+    if "ide" in vm_disk_type:
+        if destination_size < MIN_TYPE_IDE_SIZE:
+            destination_size = MIN_TYPE_IDE_SIZE
+
+    module.log(str(destination_size))
 
     # build a payload according to /rest/v1/VirDomainBlockDevice/{uuid}/clone documentation
     payload = dict(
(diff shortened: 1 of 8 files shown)


plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -242,9 +262,35 @@ def main() -> None:
...
+            #       - if keep, what could be a better name for it?
+            #           - dont_reboot_destination_vm
+            #           - keep_destination_vm_shutdown
+            #           - keep_destination_vm_off
+            #           - ignore_destination_vm_reboot --> False: reboot, True: no reboot
+            #           - ...?
+            reboot_destination_vm=dict(
+                type="bool",
+                default=True,
+            ),
```
anazobec, line 278: As this big comment explains... should I keep this parameter?

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -242,9 +262,35 @@ def main() -> None:
...
+            #       - vm_snapshot -> power_off_vm
+            #       - vm_snapshot -> power_up_vm
+            force_reboot=dict(
+                type="bool",
+                default=False,
+            ),
+            shutdown_timeout=dict(  # make this default to 300? (300 --> 5 minutes)
+                type="float",
+                default=30,
+            )
```
anazobec, line 291: Should the default be 300 seconds (as seen in the ``vm`` module) or keep it at 30?

plugins/modules/vm_snapshot_attach_disk.py
justinc1: I suggest to drop this parameter. Functionally, we get the same result if VM is shut down using `vm_params` module before `vm_snapshot_attach_disk` module is called. I do not see why we would need the extra parameter.

plugins/modules/vm_snapshot_attach_disk.py
justinc1: Also, if you search for string `scale_computing.hypercore.force_reboot` in plugins/modules, you will see we already have a documentation fragment for `force_reboot` param. Please try to reuse it.

Hopefully also code to do "nice and if needded and allowed also forced VM shutdown" can be reused.

justinc1: module_utils/vm.py, class VM, method do_shutdown_steps - this should be used. You can look at modules/vm.py, method _set_boot_order for example usage. Hope this helps.

justinc1 (commented):
just comments

plugins/modules/vm_snapshot_attach_disk.py
```diff
@@ -107,11 +107,15 @@
...
 from ..module_utils.rest_client import RestClient
 from ..module_utils.vm_snapshot import VMSnapshot
+from ..module_utils.vm import VM
 from ..module_utils.task_tag import TaskTag
 from ..module_utils.typed_classes import TypedDiff
 from typing import Tuple, Dict, Any, Optional
 
 
+# A minimum size of an IDE typed disk in HC3 API is 1 GB
+MIN_TYPE_IDE_SIZE = 1000341504  # bytes
```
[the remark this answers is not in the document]
anazobec: This is the minimum allowed size for disks of type IDE. If the size is not adjusted, then integration tests fail, because of an error (on HC3 API) when trying to reboot that VM. When I did some debugging, I found, that if I adjust the size to 1 GB (on HC3 API) by myself and then try to boot up that VM, the error is fixed...

justinc1 (commented):
x

plugins/modules/vm_snapshot_attach_disk.py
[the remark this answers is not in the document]
justinc1: This will be done in separate mini PR.

justinc1 (approved):
ty

PR #234, opened 2023-05-16
Update return of vm_snapshot_attach_disk module.

anazobec:
A small change to ``vm_snapshot_attach_disk`` was made as requested in #233.
- add ``vm_rebooted`` status to return --> info if reboot of the destination VM was performed.


justinc1:
I modified code a bit (`VM.reboot` already contained status if VM was rebooted). Extra changes were in integration tests - now we will test also what is in returned record, and if returned record does match what vm_info module returns.

justinc1 (approved):
should be  ok

PR #238, opened 2023-05-18
OIDC login

justinc1:
PR allows collection to login to HyperCore cluster using a username form external OIDC provider. New environ varibale `SC_AUTH_METHOD` (possible value None/unset, `local` or `oidc`) is added as we need to know where is user defined.

Previously we used HTTP basic login for each request. PR first refactored code to get sessionID from HC3 and reuse it for subsequent requests. 

Since OIDC is configured on .50 test cluster only, the CI job `utils_login` will be run only there.
The inventory plugin will run on .200 and .201 VSNS (just as before). But since we do not test OIDC on VSNS, OIDC is not configured there, so now we test inventory plugin on .50 also, and there (and only there) inventory plugin with OIDC user is tested.

### DIFF
--- a/.github/actions/make-integ-config/action.yml
+++ b/.github/actions/make-integ-config/action.yml
@@ -19,6 +19,9 @@ inputs:
   oidc_client_secret:
     description: 'oidc_client_secret'
     required: true
+  oidc_users_0_password:
+    description: 'oidc_users_0_password'
+    required: true
 runs:
   using: "composite"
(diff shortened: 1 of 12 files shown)


justinc1:
integ test https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5012270582

domendobnikar (approved):

PR #239, opened 2023-05-19
use ansible-core 2.15 for testing

justinc1:
Use ansible-core 2.15 for tests. Docker image for CI jobs was rebuilt and pushed to quay.io.


anazobec (approved):
I will approve this PR, since it does seem the code does indeed still work and is safe to use.

PR #240, opened 2023-05-22
Avoid invalid disk_slot in example

justinc1:
Per slack discussion we can have max 26 devices per bus. PR changed invalid disk slot in our example.


anazobec (approved):
Thank you, approved.

PR #241, opened 2023-05-22
Fix API module return list of key instead of list of dict.

anazobec:
API module with action ``get`` would sometimes return a list of with only a key present instead of dict.
This minor bug could be seen in the records output on calling the ``api`` module with action ``get`` on endpoint ``/ping`` which returned:
```
record: [
  "status"
]
```
This PR fixes this bug to by making a minor change in the ``filter_results``function in ``module_utils/utils.py``.
The return should now look like this:
```
record: [
  {
    "status": "Active"
  }
]
```

### DIFF
--- a/plugins/module_utils/utils.py
+++ b/plugins/module_utils/utils.py
@@ -13,6 +13,7 @@ import uuid
 
 from ..module_utils.errors import InvalidUuidFormatError
 from typing import Union, Any
+from copy import copy
 from ..module_utils.typed_classes import (
     TypedTaskTag,
     TypedRegistrationToAnsible,
@@ -129,7 +130,10 @@ def is_superset(superset, candidate):
 
 
 def filter_results(results, filter_data) -> list[Any]:
-    return [element for element in results if is_superset(element, filter_data)]
+    results_tmp = copy(results)
+    if type(results) is not list:
+        results_tmp = [results_tmp]
+    return [element for element in results_tmp if is_superset(element, filter_data)]
 
 
 def is_changed(
--- a/plugins/module_utils/utils.py
+++ b/plugins/module_utils/utils.py
@@ -13,7 +13,6 @@ import uuid
 
 from ..module_utils.errors import InvalidUuidFormatError
 from typing import Union, Any
-from copy import copy
 from ..module_utils.typed_classes import (
     TypedTaskTag,
     TypedRegistrationToAnsible,
@@ -130,10 +129,7 @@ def is_superset(superset, candidate):
 
 
 def filter_results(results, filter_data) -> list[Any]:
-    results_tmp = copy(results)
-    if type(results) is not list:
-        results_tmp = [results_tmp]
-    return [element for element in results_tmp if is_superset(element, filter_data)]
+    return [element for element in results if is_superset(element, filter_data)]
 
 
 def is_changed(
--- a/tests/integration/targets/api/tasks/02_corner_cases.yml
+++ b/tests/integration/targets/api/tasks/02_corner_cases.yml
@@ -13,7 +13,8 @@
 - ansible.builtin.assert:
     that:
       - ping_result is not changed
-      - ping_result.record.0.status is defined
+      - ping_result.record.status is defined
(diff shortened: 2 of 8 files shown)


tests/integration/targets/api/tasks/02_corner_cases.yml
```diff
@@ -13,7 +13,7 @@
 - ansible.builtin.assert:
     that:
       - ping_result is not changed
-      - ping_result.record.0 == "status"
+      - ping_result.record.0.status is defined
```
justinc1, line 16: I would check content of status too.

plugins/module_utils/utils.py
```diff
@@ -129,7 +130,10 @@ def is_superset(superset, candidate):
 
 
 def filter_results(results, filter_data) -> list[Any]:
-    return [element for element in results if is_superset(element, filter_data)]
+    results_tmp = copy(results)
+    if type(results) is not list:
+        results_tmp = [results_tmp]
```
justinc1, line 135: Do we really want to wrap plain dict into list? It seems more logical for api module to just return plain response from HyperCore API.

justinc1 (commented):
I think api module should just return what API does return. I'm note sure how much such "requirement" complicates the work.

anazobec:
Yes, I've mentioned that some integration tests fail. Before the latest commit, 12 tests failed. 9 with error ``Connection Refused`` and others with just an assertion failure.

#### Integ tests with Connection Refused
- vm_clone
- virtual_disk_attach
- virtual_disk
- version_update__shutdown_restart
- git_issue_3
- git_issue_15
- git_issue_11
- certificate (both of them)

#### Integ tests with Assertion failure
- version_update_info
- git_issue_41
- utils_login

The results of these integration tests, can be seen from here: https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5043144438/jobs/9044620677

anazobec:
With the latest commit, API module should now on ``/ping`` endpoint, with data ``{ status: "Active" }`` return something like this:
```
record: {
  "status": "Active"
}
```

tests/unit/plugins/module_utils/test_rest_client.py
```diff
@@ -64,6 +64,42 @@ def test_query_passing(self, client):
...
+            200, '{"result": []}', {"X-Total-Count": "0"}
+        )
+        t = rest_client.RestClient(client)
+
+        t.list_records_raw("my_table")
+
+        client.get.assert_called_once_with(
+            path="my_table",
+            timeout=None,
+        )
```
justinc1: This one is same as `test_empty_response`, only `list_records_raw` return value is not checked. We should drop the duplicated unittest.

anazobec:
Here are the newly run integration tests as of commit [d1e799e](https://github.com/ScaleComputing/HyperCoreAnsibleCollection/pull/241/commits/d1e799ec29714115d33ae876b726cfac22354b15): https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5045862848

11 tests fail: 2 are assertion errors, 1 of them is an error caused by changing the ``get_records`` function in the api module. The rest of the tests fail with a Connection Refused error, Unexpected response, EOF.

#### Assertion failure:
1. version_update_info (.200)
2. utils_login (.50) => is caused by fixing api module

#### Connection Refused failure:
1. version_update__shutdown_restart_vms (.200)
2. email_alert (.200)
3. git_issue_15 (.201)
4. git_issue_11 (.201)
5. api (.201)
6. virtual_disk (.201)

#### Other failures:
1. certificate (.200, .201) => EOF in violation of protocol (_ssl.c: 997)
2. git_issue_41 (.200) => Unexpected response - 500

justinc1 (approved):
Thank you, nice work. I did very trivial changes, just small unittest/comment cleanup.

tests/unit/plugins/module_utils/test_rest_client.py
anazobec: Thank you.

PR #242, opened 2023-05-23
Bump expected latest released HyperCore version

justinc1:
In tests we have hardcoded expected HyperCore version. The 9.2.17 is GA (general availability), so this is what our test cluster should report as next available version.

I hate we need to manually bump this string every few weeks. Maybe we should have more complex integration test that could check not only "next_version == 9.2.17.211525", but also "next_version >= 9.2.17.211525".

### DIFF
--- a/examples/version_update_single_node.yml
+++ b/examples/version_update_single_node.yml
@@ -7,7 +7,7 @@
   check_mode: true
 
   vars:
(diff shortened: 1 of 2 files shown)


anazobec (approved):
Looks approvable so, approving. :+1:

PR #252, opened 2023-07-03
Fix vm disk remove

justinc1:
Sometimes we can remove disk from a running VM, and sometimes we need to shutdown VM before disk can be removed. It depends on HyperCore thinking VM might be using the disk.

The PR:
- first tries to remove disk from a running VM (the behaviour before this PR)
- if this fails, it shutdowns VM, tries delete a second time and start VM back

Integration tests for all this is included, for `vm` and `vm_disk` modules. It covers 3 cases:
- VM being off
- VM running and cannot be shutdown (it does not respond to nice shutdown, and force shutdown is not allowed)
- VM running and can be shutdown (force shutdown is allowed)

A small bug in vm_disk module was also fixed. The module didn't start back VM if it was shutdown during execution.

### DIFF
--- a/plugins/module_utils/errors.py
+++ b/plugins/module_utils/errors.py
@@ -7,7 +7,8 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
-from typing import Union
+import json
+from typing import Union, Dict, Any
 from ansible.module_utils.urls import Request
 
 
@@ -111,3 +112,14 @@ class ScaleTimeoutError(ScaleComputingError):
     def __init__(self, data: Union[str, Exception]):
         self.message = f"Request timed out: {data}."
         super(ScaleTimeoutError, self).__init__(self.message)
+
+
+class TaskTagError(ScaleComputingError):
+    def __init__(self, task_status: Dict[Any, Any]):
+        # task_status is dict returned by GET /rest/v1/TaskTag
+        message = "There was a problem during this task execution."
+        message += f" Task details: {json.dumps(task_status)}"
+        self.message = message
+        self.task_status_state = task_status["state"]
+        self.task_status = task_status
+        super().__init__(self.message)
--- a/plugins/module_utils/vm.py
+++ b/plugins/module_utils/vm.py
@@ -1091,6 +1091,10 @@ class ManageVMDisks:
                 ):
                     to_delete = False
             if to_delete:
+                # HyperCore is sometimes able to delete disk on running VM,
+                # but sometimes we need to shutdown VM to remove disk.
+                # It is hard to know in advance if shutdown is required.
+                # We try to remove disk without shutdown, if delete fails, we shutdown VM and try again.
                 if existing_disk.needs_reboot("delete"):
                     vm.do_shutdown_steps(module, rest_client)
                 task_tag = rest_client.delete_record(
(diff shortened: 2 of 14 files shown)


justinc1:
Integ test: https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5442362156

plugins/module_utils/vm.py
```diff
@@ -1206,13 +1236,15 @@ def ensure_present_or_set(cls, module, rest_client, module_path):
...
+            vm_before.vm_power_up(module, rest_client)
             vm_after, disks_after = cls.get_vm_by_name(module, rest_client)
             return (
                 changed,
                 disks_after,
                 dict(before=disks_before, after=disks_after),
                 vm_before.reboot,
             )
+        vm_before.vm_power_up(module, rest_client)
         return changed, vm_before.reboot
```
shoriminimoe, line 1248: I think it would be more clear to put the `vm_power_up` call before line 1238 and remove the other one. The call is made whether the `if` condition is satisfied or not.

```suggestion
        vm_before.vm_power_up(module, rest_client)
        if called_from_vm_disk:
            vm_after, disks_after = cls.get_vm_by_name(module, rest_client)
            return (
                changed,
                disks_after,
                dict(before=disks_before, after=disks_after),
                vm_before.reboot,
            )
        return changed, vm_before.reboot
```

plugins/module_utils/errors.py
```diff
@@ -111,3 +112,14 @@ class ScaleTimeoutError(ScaleComputingError):
     def __init__(self, data: Union[str, Exception]):
         self.message = f"Request timed out: {data}."
         super(ScaleTimeoutError, self).__init__(self.message)
+
+
+class TaskTagError(ScaleComputingError):
```
shoriminimoe, line 117: This is great. Raising this error type is clear and the message is useful. :+1:

plugins/module_utils/vm.py
```diff
@@ -1099,7 +1103,33 @@ def _delete_not_used_disks(cls, module, rest_client, vm, changed, disk_key):
...
+                        != "Unable to delete block device from VM '%@': Still in use"
+                    ):
+                        raise
+                    vm_fresh_data = rest_client.get_record(
+                        f"/rest/v1/VirDomain/{vm.uuid}", must_exist=True
+                    )
+                    if vm_fresh_data["state"] != "RUNNING":
+                        raise
+                    # shutdown and retry remove
+                    vm.do_shutdown_steps(module, rest_client)
```
shoriminimoe, line 1125: I'm not clear on the full context for this call. Is the user aware that the VM may be shutdown by attempting to delete a disk? Would it be better to leave this in the user's hands? In other words, if the VM needs to be shutdown before deleting the disk, should the user be required to include that in their playbooks? 

I see the previously existing call to shutdown on line 1099. That implies to me that the shutdown is OK here but I wanted to ask the question. Is the shutdown call only honored if `force_reboot: True` or something?

shoriminimoe (approved):
I'm still learning Ansible testing so this may be an ignorant question. Are the new `vm__*` and `vm_disk__*` tests not the same thing? What is the difference?

.github/workflows/integ-test.yml
```diff
@@ -208,6 +209,7 @@ jobs:
       run:
         working-directory: ${{ env.WORKDIR }}
     strategy:
+      max-parallel: 1
```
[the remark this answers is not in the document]
justinc1: I think this would not result in each sc_host running at most 1 test. We provide a matrix with "sc_host, test_name" fields. To github sc_host is not more important than test_name.

But something like this might work.

```
concurrency:
  group: ci-${{ matrix.sc_host }}
```

justinc1:
> Are the new vm__* and vm_disk__* tests not the same thing?

They are very similar, and both modules use same code from module_utils. But they are not exactly the same, so I added separate tests (and after doing this, I noticed vm_disk module does not correctly restart VMs).

plugins/module_utils/vm.py
justinc1: User is aware module will try to shutdown VM if this is required to remove a disk: Docs:
```
  - Module tries to remove disks from a running VM.
    If disk cannot be removed from running VM,
    then VM will be shutdown, disk will be removed, and VM is started back.
  - VM has C(shutdown_timeout) time to respond to shutdown request.
    If VM is not shutoff within I(shutdown_timeout),
    then a force shutdown will be issued if C(force_reboot=True).
```

The `force_reboot` could be called `allow_force_reboot_if_nice_shutdown_does_not_work`.

.github/workflows/integ-test.yml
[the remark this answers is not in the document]
shoriminimoe: Good call. I thought the order you declare the matrix values defined the execution order.

PolonaM (approved):
Thanks Justin for the work! Except for the mentioned replacement of task_tag with task_status and maybe even adding the task status to typed_classes.py, everything else looks good to me.

PR #253, opened 2023-07-07
Fix vm clone test

justinc1:
The vm_clone works correctly. The test however assumed vm_snapshot_info output is sorted. But it is not. PR now sorts output from vm_snapshot_info before using one of two snapshots with `label=="test2"`.

Test is `14_clone_from_snapshot.yml` a bit modified, each of 3 snapshots has a different CPU and disk count. Then it is easier to check that correct snapshot was used to create a new VM.


justinc1:
Integ test: https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/5486162516

PolonaM (approved):

PR #256, opened 2023-07-07
ci: add publish-docs job to release workflow

shoriminimoe:


justinc1:
Nice and ty.

I suggest we also stop building/deploying docs for gh-pages branch. We do not need this extra branch at all, so I already removed it.

shoriminimoe:
> Nice and ty.
> 
> I suggest we also stop building/deploying docs for gh-pages branch. We do not need this extra branch at all, so I already removed it.

Sounds good to me. I will remove that trigger from the workflow file. Are there any other uses of the `gh-pages` branch in the repo that also need to be updated?

justinc1:
gh-pages branch was created with idea to:
1. build docs from main branch, than
2. commit built docs into gh-pages branch, then
3. deploy gh-pages branch as gh-pages.

But the 2nd step is not needed at all, we can deploy directly from main branch or git tag.

shoriminimoe:
@justinc1 I rebased this onto the latest `main` commit. Is there anything else needed for it?

justinc1 (approved):
Thank you. Nothing else is needed.

PR #266, opened 2023-09-26
Small fixes

justinc1:
Only small fixes. The debug HTTP request logging can be enabled with `SC_DEBUG_LOG_TRAFFIC=1` environ variable.  The code crashed if request timed out, as we wanted to log HTTP response, but variable was not set.

The CI test need to expect 9.2.17.211525 as available update (HyperCore version upgrade).

The cleanup on SMB server is modified to run even if tests fail. The cleanup will remove all files not from current month (before: not from current day). This cloud interfere with tests on first/last day of month, but in practice we should not have a problem as the cleanup will be scheduled before tests.

Similar for DR target cluster, we will remove all VMs older that 2 days. Should not interfere with tests at all.
Here we remove all old replicated VMs with tag "Xlab" (before tag Xlab and CI was needed; but we forgot to include "CI" string in some tests, so VMs were not removed).

### DIFF
--- a/.github/workflows/integ-test.yml
+++ b/.github/workflows/integ-test.yml
@@ -366,9 +366,7 @@ jobs:
       - run: ansible-playbook tests/integration/cleanup/ci_replica_cleanup.yml
 
   smb_cleanup:
-    needs:
-      - integ
-      - integ-seq
+    needs: []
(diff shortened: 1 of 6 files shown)


domendobnikar (approved):
LGTM :+1:

PR #268, opened 2023-09-28
Ci for 9.3

justinc1:
We were missing wait on tasktag for SnapshotSchedule create, and wait on object begin delete for SnapshotSchedule and VirtualDisk.

Other changes are just small CI speedup, and fix how CI test check if HyperCore version >= 9.2.

### DIFF
--- a/plugins/modules/snapshot_schedule.py
+++ b/plugins/modules/snapshot_schedule.py
@@ -128,11 +128,13 @@ record:
 """
 
 from ansible.module_utils.basic import AnsibleModule
+import time
 
 from ..module_utils import arguments, errors
 from ..module_utils.client import Client
 from ..module_utils.rest_client import RestClient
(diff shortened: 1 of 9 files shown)


justinc1:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/6349507160
 - 14 failed jobs after 1st try
 - ? failed jobs after 2nd try

shoriminimoe (approved):
I'm mixed about adding the sleeps. Would it instead be better to rely on the user (and the tests) to use ansible features like [`until`](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met) in their tasks in the event of request failures?

justinc1:
The `until` would be used in playbook. But I wanted to hide this extra delay into the module; user should not need to know it takes a few ms for object deletion.

PR #278, opened 2023-11-28
Ci ansible core 2.16

justinc1:
Add ansible-core 2.16 to test matrix. Also new CI image was build and pushed to quay.io

Small style changes were needed for ansible sanity test to pass.

### DIFF
--- a/ci-infra/docker-image/build.sh
+++ b/ci-infra/docker-image/build.sh
@@ -11,7 +11,7 @@ set -eux
 # Where to push images
 DOCKER_REGISTRY_REPO=quay.io/justinc1_github/scale_ci_integ
 # Tag to push
-DOCKER_IMAGE_TAG=8
+DOCKER_IMAGE_TAG=9
 
 DOCKER_CACHE="${DOCKER_CACHE:-n}"
 if [ "$DOCKER_CACHE" == "n" ]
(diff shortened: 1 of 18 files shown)


plugins/module_utils/client.py
```diff
@@ -215,13 +215,14 @@ def _request_no_log(
             if (
                 e.args
                 and isinstance(e.args, tuple)
-                and type(e.args[0]) == ConnectionRefusedError
+                # and type(e.args[0]) == ConnectionRefusedError
```
shoriminimoe, line 218: I would just remove this line :man_shrugging:

```suggestion
```

ci-infra/docker-image/build.sh
```diff
@@ -27,7 +27,7 @@ REPO_DIR="$(realpath "$THIS_SCRIPT_DIR/../.." )"
 cat "$REPO_DIR"/{docs,test,sanity,mypy}.requirements > all.requirements
 # TODO but is ghrc.io feature available/enabled?
 # Or push to quay.io?
-docker build "$DOCKER_BUILD_OPT" -t "$DOCKER_REGISTRY_REPO:$DOCKER_IMAGE_TAG" .
+docker build $DOCKER_BUILD_OPT -t "$DOCKER_REGISTRY_REPO:$DOCKER_IMAGE_TAG" .
```
shoriminimoe, line 30: Was removing the quotes here intentional? I usually keep variables quoted when I use them as command arguments to avoid word splitting, but it looks like that won't matter here. Just checking.

shoriminimoe (approved):
Looks good.

ci-infra/docker-image/build.sh
justinc1: Removing quotes was required. A bit older docker-ce client was happy with empty DOCKER_BUILD_OPT - it just ignored "". But most recent one did complain.

PR #308, opened 2024-04-04
Include tests/config.yml into tar.gz

justinc1:
Having `tests/config.yml` in built `.tar.gz` should help `ansible-test sanity` to not test with old python version (2.7 etc).


shoriminimoe (approved):
:rocket:

PR #310, opened 2024-05-08
url2template and template2vm roles

justinc1:
2 new roles are added.

url2template creates template VM from disk image URL.
Example URL is https://cloud-images.ubuntu.com/releases/22.04/release-20240416/ubuntu-22.04-server-cloudimg-amd64.img. 

Next template2url can be used to create new VMs from template VMs. Here we typically need to configure VM with cloud-init.
Cloud-init configuration data depends on the used image. A working example for above Ubuntu image is included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ tests/integration/inventory
 
 docs/build
 docs/source/modules
+ci-infra/local-dev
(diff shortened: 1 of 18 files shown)


domendobnikar (approved):
Looks good, thank you :+1:

PR #318, opened 2024-05-22
Ci refactor

justinc1:
Running more than one CI test per HyperCore cluster is a bit problematic. PR refactor `integ-test` job to same pattern as we have for `integ-seq-test` - include the helper workflow `z_ansible-test.yml`. This allows running tests on all 4 clusters in parallel, and at most 1 test pre cluster.

Some tests require physical cluster. Those are run in newly added `integ-physical` job, on https://10.5.11.50.

We test examples with simple `ansible-playbook ... example_name.yml`. This does not work for some examples, so do not test those to avoid false failures.

### DIFF
--- a/.github/job_retry_needed.py
+++ b/.github/job_retry_needed.py
@@ -90,6 +90,19 @@ def main():
             job["id"], job["status"], job["conclusion"], job["name"]
         )
 
+    # Double check input job name is one of actual job names
+    # Naming convention changed at some time:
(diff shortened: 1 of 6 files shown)


justinc1:
Integ test: https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/9194109175

shoriminimoe (approved):
This is so much better! Thank you! :clap:

PR #322, opened 2024-07-02
Fail late if cluster_config task fails

### LINKED ISSUE #321: :lady_beetle: Bug: cluster_config role stops if NTP can't be validated
## Describe the bug

**To Reproduce**
cluster_config role will stop setting other values if time_server: fails to set 

in this case because of invalid dns (expected for this "staging" environment) preventing us from resolving IP 

TASK [scale_computing.hypercore.cluster_config : Set time server] ************************************************************************************************************************
Wednesday 26 June 2024  12:22:50 -0400 (0:00:00.017)       0:00:02.597 ******** 
fatal: [staging4]: FAILED! => {"changed": false, "msg": "Unexpected response - 400 b'{\"error\":\"pool.ntp.org is not a valid NTP server\"}'"}

## Expected behavior

I think it would be expected to continue setting other parameters

## System Info (please complete the following information):
 - HyperCore Version: 
 - Ansible Version: 
 - Collection Version [e.g. 22]

justinc1:
If cluster_config role fails at some task, it should continue with other tasks - this was requested in #321.

The email_alerts and syslog_servers needs to be updated to tolerate partial failure. I plan to do this in separate MRs.

Partially fixes #321.


tests/integration/targets/role_cluster_config/tasks/01_a.yml
domendobnikar, line 1: Suggestion: rename the _a and _b file to something that gives a little bit more information what is actually in the file

domendobnikar (approved):
Left a suggestion otherwise looks good, thank you 👍

tests/integration/targets/role_cluster_config/tasks/01_a.yml
justinc1: The basics (tm). Did rename to "01_cluster_config_a.yml", as we apply "cluster_config_a" configuration set.

PR #324, opened 2024-07-11
Syslog server set

justinc1:
Extend `syslog_server` module with `state=set` option. This allows to configure HyperCore to exactly match given Ansible configuration.

The `syslog_server` module output is extended with `records` field, as this now makes more sense.

A corner case bug in `syslog_server` module was also fixed. Before code crashed if we tried to both rename and modify (say change port), and object was already renamed.


plugins/modules/syslog_server.py
```diff
@@ -242,82 +290,189 @@ def update_syslog_server(
...
+    records_after = SyslogServer.get_state(rest_client)
     return (
         True,
         {},
+        records_after,
         dict(before=before, after={}),
     )  # changed, records, diff
 
 
+def _are_equal(
```
domendobnikar, line 328: Should this be an `__eq__ ` method and part of the `syslogserver ` class?

domendobnikar (approved):
Left a comment otherwise looks good, thank you 👍

plugins/modules/syslog_server.py
justinc1: Yes. No. Both.

Class `SyslogServer` already has an `__eq__` method, with comment `# This method is here for testing purposes!` - the method returns `True` only if all attributes of `self` and `other` are equal. Also attributes `uuid` and `alert_tag_uuid` are checked. So `True` means not only the configured values are same, but both objects are the same (uuid is same too).

 Here I needed to compare only attributes that can be changed via GUI/ansible module. As many of our modules already have `__eq__` with same "for CI testing only" comment, I decided to leave `__eq__` unmodified.

I will move this to `SyslogServer.is_equivalent`, with comment why this in not `__eq__`.

PR #330, opened 2024-08-27
ci: replace VSNS 9.4.0 -> 9.4.17

shoriminimoe:


justinc1 (approved):
All OK, only `current_version: "9.4.17.215487"` needs to be updated. I will puhs a separate commit.

shoriminimoe:
> All OK, only `current_version: "9.4.17.215487"` needs to be updated. I will puhs a separate commit.

Good catch. Thank you!

PR #331, opened 2024-08-28
ci: document NTP server, SMB server and OIDC

justinc1:
The github runner was configured back to IP it had before reboot (10.5.11.5). `DEVELOPMENT.md` is extended to mention NTP, SMB and OIDC setup we use in CI testing.


shoriminimoe (approved):
This is helpful information. 🚀

PR #337, opened 2024-09-10
Fix docs build

justinc1:
Generate HTML documentation was wrong. All modules contained something like `extra fields not permitted (type=value_error.extra)`. Problem was we were using outdated antsibull-docs. The new docker image uses latest antsibull-docs.

HTML docs built by CI job are at https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/10793721063/artifacts/1914436268.


shoriminimoe (approved):
Thanks @justinc1!

PR #347, opened 2025-01-06
Document cloud-init is applied only to newly created VM

### LINKED ISSUE #340: :lady_beetle: Bug: vm module doesn't set send cloud_init for existing VMs
## Describe the bug
The `vm` module has a `cloud_init` parameter which works as described when creating new VMs. However when applied to an existing VM, the supplied `cloud_init` parameter is silently ignored. 

**To Reproduce**
1. Run the `vm` module on an existing VM and provide any `cloud_init` parameters.
2. Observe that the task does not error and the VM doesn't have a cloud init iso attached afterwards.

## Expected behavior
Either the documentation for the `vm` module should state that the `cloud_init` parameter only applies to newly created VMs or the module should apply any `cloud_init` parameter to existing VMs. 

## System Info (please complete the following information):
 - OS: Linux
 - HyperCore Version: 9.4.18.215718
 - Ansible Version: core 2.14.14
 - Collection Version: 1.6.0

## Additional context
There doesn't seem to be a way to attach cloud init iso to an existing VM elsewhere in the collection, besides doing it manually with the `api` module.

justinc1:
Fixes #340

There was some confusion about why cloud-init ISO was not added to existing VM. Documentation should prevent confusion.

### DIFF
--- a/plugins/doc_fragments/cloud_init.py
+++ b/plugins/doc_fragments/cloud_init.py
@@ -16,8 +16,12 @@ options:
   cloud_init:
     description:
       - Configuration to be used by cloud-init (Linux) or cloudbase-init (Windows).
-      - When non-empty will create an extra ISO device attached to VirDomain as a NoCloud datasource.
+      - When non-empty will create an extra ISO device attached to VM as a NoCloud datasource.
       - There has to be cloud-config comment present at the beginning of cloud_init file or raw yaml.
+      - The cloud-init configuration is applied only during the initial creation of a VM
+        (whether through provisioning, cloning, or importing).
+        It is not applied to an existing VM.
+        This is consistent with the behavior of the HyperCore appliance.
     required: false
     type: dict
     default: {}


ddemlow:
not sure where this will show up - but will it be clear that includes "from scratch" VMs as well as cloned VMs and vm's created via VM import?

The cloud-init configuration is applied only to newly created VMs.

justinc1:
@ddemlow Would be this text better?

```
      - The cloud-init configuration is applied only to a newly created VM.
        It is not applied to an existing VM, to a cloned VM, or to an imported VM.
        This is consistent with HyperCore appliance behavior.
```

ddemlow:
it IS applied when VMs are cloned or imported as well  - well when those result in new vms being created ... which I guess is whenever the target vm doesn't already exist?   

for example:  
https://scalecomputing.github.io/HyperCoreAnsibleCollection/collections/scale_computing/hypercore/vm_clone_module.html#parameter-cloud_init

how about this:

	•	The cloud-init configuration is applied only during the initial creation of a VM (whether through provisioning, cloning, or importing).
It is not applied to an existing VM.
This is consistent with the behavior of the HyperCore appliance.

justinc1:
Aha. Then it is up to the new (created, cloned or imported) VM to decide what to do with cloud-init configuration. I mean, some cloud-init configurations are applied only on "first run". Such commands would not be applied when VM is cloned or imported - the cloud-init daemon already run before.

Thank you for corrected doc snippet, Dave! Will use it.

ddemlow:
correct.... our api will only create and attach the cidata iso if the vm is considered new through any of those methods ... whether the guest image is "armed" to look for that and use / apply  it is an entirely different story

PR #351, opened 2025-01-31
Extend vm_disk module tests - assert on vm_info output

justinc1:
Using `vm_info` means we test what `vm_disk` actually changed on HyperCore cluster, not only what `vm_disk` return value (e.g. what `vm_disk` thinks it applied).


justinc1:
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/13067636820 passed on all HyperCore versions

PR #359, opened 2025-06-02
Changed regenerate_disk_id default to false

### LINKED ISSUE #358: :rocket: Feature request: Change Virtual disk parameter regenerate_disk_id default to False
## Is your feature request related to a problem? Please describe.
We had a discussion about Virtual Disk parameters and their default values differences between Terraform and Ansible collections. We came to an agreement that both collections should use the same default values for the same parameters.

## Describe the solution you'd like
Change the regenerate_disk_id default value to False.

domendobnikar:
What was done in this PR:
- Changed regenerate_disk_id default to False as agreed to be in sync with what was done for Terraform Plugin


justinc1 (approved):
TY

PR #362, opened 2025-09-08
Easier local integration testing, SMB server 

justinc1:
PR introduces a few small changes for easier local testing.

A pylint was complaining.

A `ci-infra/helpers/run-tests.sh` makes it easier to run and retry only failed integration tests.
It is intended for testing locally, on developers laptop. usage:
- On first run, all tests (or only those listed in `test-names.txt`) are run.
- Test result in stored in `test-names.txt`.
- On second run, only tests marked as failed in `test-names.txt` are retried.

A docker compose to setup SMB server for VM import/export testing is added. With this, I was able to check `vm_import` and `vm_export` modules do work.

A shutdown timeout for vm_snapshot_attach_disk is adjusted, to make test faster.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -10,3 +10,4 @@ exclude_paths:
   - .github/workflows
   - changelogs
   - .tox
+  - ci-infra/smb-server/compose.yml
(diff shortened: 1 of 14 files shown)


anazobec (approved):
Thanks for the improvement, @justinc1! LGTM

PR #363, opened 2025-09-08
Logout when module finishes

justinc1:
Logout from HyperCore when module finishes. Otherwise HyperCore needs to auto-close oldest connections (the least recently used one), and this sometimes selects "wrong" connection. For example, while running integration tests, the browser session might get logged out.


anazobec (approved):
Great! Implementation looks good, thanks @justinc1!

PR #364, opened 2025-09-22
Support HyperCore version update to any version

### LINKED ISSUE #361: :rocket: Feature request:  test / allow ansible to request update for any specific build UUID
## Is your feature request related to a problem? Please describe.
sometimes I need to update a cluster to a release that is not yet widely available (but I know the specific build I want to update to) 

## Describe the solution you'd like

I learned / confirmed today that in 9.6.?  we not only added the ability for the UI to request any specific hcos build that we've pushed externally  (even if it doesn't show up as "available" in UI) image.png

the api will also let you request that 

curl -X 'POST' \
  'https://192.168.1.230/rest/v1/Update/9.6.7.220258/apply' \
  -H 'accept: application/json' \
  -d ''

I am curious if ansible module would already allow that or if changes would needed to be made?  in this case that 9.6.7.220258/ would not show up in a get?  wasn't sure if plugin checks or not?

justinc1:
Support HyperCore version update to any version, even if update is not listed under "available updates". An additional `force_update` (or `version_update_single_node_force_update`) param is added to allow this.

Fix #361

### DIFF
--- a/changelogs/fragments/20250922-update-hypercore-to-any-version.yml
+++ b/changelogs/fragments/20250922-update-hypercore-to-any-version.yml
@@ -1,0 +1,4 @@
+---
+major_changes:
+  - Support HyperCore update to arbitrary version, even if it is not listed under available updates.
+    (https://github.com/ScaleComputing/HyperCoreAnsibleCollection/pull/364)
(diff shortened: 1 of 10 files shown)


anazobec:
Code looks good to me. Thanks @justinc1! Approving.

anazobec (approved):

PR #365, opened 2025-09-26
CI use GHA from ansible-network/github_actions

justinc1:
PR wants to use same linters as https://github.com/ansible-collections/amazon.aws uses.
This should make it easy to test with correct ansible-lint version etc.

Some mechanical code changes were needed.


anazobec (approved):
Code looks good from my side. Thanks @justinc1!

PR #371, opened 2026-06-03
370 - vm_clone produces VMs that are silently fragile to multi-disk attach

### LINKED ISSUE #370: vm_clone produces VMs that are silently fragile to multi-disk attach + reboot (no bootDevices, vda-anchored grub)
## Summary

VMs cloned via `scale_computing.hypercore.vm_clone` from a standard Canonical Ubuntu cloud image inherit two independent boot-fragility issues that surface the first time a second `VIRTIO_DISK` is attached AND the VM is rebooted:

1. **`bootDevices` is empty** on the cloned VirDomain. HC's BIOS has an implicit auto-fallback to "the only virtio disk" but that fallback gives up once a second virtio is attached → `Boot failed: not a bootable disk / No bootable device`.
2. **The cloud image's `/etc/default/grub` ships with `GRUB_DISABLE_LINUX_UUID=true`**, so the generated `/boot/grub/grub.cfg` uses `root=/dev/vda1` (device-path) instead of `root=UUID=…` (UUID-based). When a second virtio disk reorders PCI enumeration, the OS disk becomes `vdb`, the kernel can't find `/dev/vda1`, and initramfs hangs at `Btrfs loaded …`.

Combined effect: a cloud-image VM cloned via this module **boots correctly for years** while it has exactly one virtio disk, then bricks on the first reboot after any tool (CSI driver, manual disk-add, Terraform provider, or even another playbook in the same suite) attaches a second virtio disk.

This is the underlying root cause for a pair of related issues filed in `ScaleComputing/k3s-on-hypercore` (#7 and #8), but the same exposure exists in **every** downstream consumer of `vm_clone`: `k3s-ansible-hypercore`, `ansible_edge_playbooks`, and customer playbooks that clone from cloud-image templates.

## Reproducer

```bash
# Clone an Ubuntu cloud-image VM via this collection (any of the standard
# patterns, e.g. ansible_edge_playbooks/simple_vm_deploy.yml):
ansible-playbook simple_vm_deploy.yml

# Inspect the resulting VM:
curl -sk -u admin:admin "https://<hc-host>/rest/v1/VirDomain/<new-vm-uuid>" \
  | jq '.[0] | {name, bootDevices}'
# → bootDevices: []   ← issue #1

# SSH into the new VM:
grep GRUB_DISABLE_LINUX_UUID /etc/default/grub
# → GRUB_DISABLE_LINUX_UUID=true     ← issue #2
cat /proc/cmdline
# → root=/dev/vda1    ← issue #2 consequence

# Attach a second VIRTIO_DISK, then reboot:
#   - If issue #1 unaddressed: BIOS "No bootable device" — VM unbootable.
#   - If issue #1 fixed but #2 unaddressed: kernel loads but initramfs hangs.
```

### Layer 1: `vm_clone` should set `bootDevices` after clone

The most natural fix is to add a parameter (or change defaults) so the cloned VM's `bootDevices` is populated with the primary disk's UUID immediately after creation. Something like:

```yaml
- name: Clone the VM
  scale_computing.hypercore.vm_clone:
    vm_name: "{{ inventory_hostname }}"
    source_vm_name: "{{ source_template }}"
    set_boot_devices: yes   # new default-yes parameter
```

Or simpler: always populate `bootDevices` with the cloned VM's largest VIRTIO_DISK by default. Users who want to manage boot order manually can override.

### Layer 2: optional `cloud_init.runcmd` injection

If the module accepts `cloud_init.user_data`, the generated cloud-init should include a `runcmd` to fix the grub UUID issue at first boot:

```yaml
runcmd:
  - sed -i 's/^GRUB_DISABLE_LINUX_UUID=true/#GRUB_DISABLE_LINUX_UUID=true/' /etc/default/grub
  - update-grub
```

Either bake this into the module's default user_data, or document it prominently in the role's README so every downstream playbook can add it.

## Workaround for the field today

I've documented both layers + per-VM fix recipes in an internal `hypercore-api-notes.md` reference. The TL;DR for any playbook today:

1. After `vm_clone`, `PATCH /VirDomain/{uuid}` with `bootDevices: [<primary-disk-uuid>]`.
2. Include the grub `runcmd` snippet in the cloud-init user data.

I've also done both layers manually on an 8-VM reference cluster — works cleanly.

## Why this matters

Without these fixes, **every k8s-on-HyperCore deployment is a ticking time bomb**. The VMs come up fine, work fine for months/years, then the first time someone adds storage (CSI driver, data disk, NFS export disk, etc.) and the VM reboots, it bricks. That's exactly the wrong failure pattern: silent and recovery-blocking.

## Related

- HC platform team issue (BIOS auto-fallback should be smarter or refuse empty `bootDevices`): `github.lab.local/dev/dev#394`
- `k3s-on-hypercore#7` (bootDevices)
- `k3s-on-hypercore#8` (grub UUID)

## Suggested labels

`bug`, `priority:high` — silent failure mode that affects every multi-disk workload on HC.

domendobnikar:
Boot order is now set after cloning to avoid issues when attaching additional disk to VM afterwards. 
`Cloud init's user data` default `runcmd `is also added if `cloud init` has been passed as parameter.

### DIFF
--- a/plugins/module_utils/vm.py
+++ b/plugins/module_utils/vm.py
@@ -218,7 +218,7 @@ class VM(PayloadMapper):
         power_state=None,
         power_action=None,
         nics=None,  # nics represents a list of type Nic
-        disks=None,  # disks represents a list of type Nic
+        disks=None,  # disks represents a list of type Disk
         # boot_devices are stored as list of nics and/or disks internally.
         boot_devices=None,
         attach_guest_tools_iso=False,
@@ -423,6 +423,27 @@ class VM(PayloadMapper):
             payload["template"]["cloudInitData"] = cloud_init
         return payload
 
+    @staticmethod
+    def clone_add_user_data_to_cloud_init(cloud_init):
+        # Task 370 - the generated cloud-init should include a runcmd
+        # to fix the grub UUID issue at first boot
+
+        user_data = cloud_init.get("user_data")
+        if not user_data:
+            return cloud_init
+
+        cloud_init["user_data"] = (
+            user_data.rstrip()
+            + """
+
+        runcmd:
+        - sed -i 's/^GRUB_DISABLE_LINUX_UUID=true/#GRUB_DISABLE_LINUX_UUID=true/' /etc/default/grub
+        - update-grub
+        """
+        )
+
+        return cloud_init
+
     @classmethod
     def create_clone_vm_payload(
         cls,
@@ -446,6 +467,7 @@ class VM(PayloadMapper):
                     hypercore_tags.append(tag)
             data["template"]["tags"] = ",".join(hypercore_tags)
         if cloud_init:
+            cloud_init = cls.clone_add_user_data_to_cloud_init(cloud_init)
             data["template"]["cloudInitData"] = cloud_init
         if preserve_mac_address:
             data["template"]["netDevs"] = [
@@ -609,6 +631,13 @@ class VM(PayloadMapper):
         for disk in self.disk_list:
             if disk.slot == slot:
                 return disk
+
+    # primary disk is the largest Virtio disk
+    def get_primary_disk(self):
+        virtio_disks = [disk for disk in self.disk_list if disk.type == "virtio_disk"]
+        if not virtio_disks:
+            return None
+        return max(virtio_disks, key=lambda disk: disk.size)
 
     def post_vm_payload(self, rest_client, ansible_dict):
         # The rest of the keys from VM_PAYLOAD_KEYS will get set properly automatically
--- a/plugins/modules/vm_clone.py
+++ b/plugins/modules/vm_clone.py
@@ -166,10 +166,29 @@ def run(module, rest_client):
     TaskTag.wait_task(rest_client, task)
     task_status = TaskTag.get_task_status(rest_client, task)
     if task_status and task_status.get("state", "") == "COMPLETE":
-        return (
-            True,
-            f"Virtual machine - {module.params['source_vm_name']} - cloning complete to - {module.params['vm_name']}.",
-        )
+        # Get cloned VM
+        virtual_machine_cloned_obj = VM.get_or_fail(query={"name": module.params["vm_name"]}, rest_client=rest_client)[
+            0
+        ]
+        # Set boot devices after cloning Issue-370 (VM starts failing as soon as another disk is attahed if boot is not specified)
+        # By default we always set the largest Virtio disk which is the "primary disk"
+        primary_disk = virtual_machine_cloned_obj.get_primary_disk()
+        boot_items = [primary_disk.to_ansible()] if primary_disk else []
+        # previous boot order after cloning is always empty
+        previous_boot_order = []
+        changed = virtual_machine_cloned_obj.set_boot_devices(boot_items, module, rest_client, previous_boot_order)
+        if changed:
+            msg = "and boot order was set - you can change it with vm_boot_devices module"
+            return (
+                True,
+                f"Virtual machine - {module.params['source_vm_name']} - cloning complete to - {module.params['vm_name']} {msg}.",
+            )
+        else:
+            msg = "and boot order was not set - you can set it with vm_boot_devices module"
+            return (
+                True,
+                f"Virtual machine - {module.params['source_vm_name']} - cloning complete to - {module.params['vm_name']} {msg}.",
+            )
     raise errors.ScaleComputingError(
         f"There was a problem during cloning of {module.params['source_vm_name']}, cloning failed."
     )
--- a/tests/integration/targets/vm_clone/tasks/10_clone_stopped.yml
+++ b/tests/integration/targets/vm_clone/tasks/10_clone_stopped.yml
@@ -29,11 +29,12 @@
       - cloned_info.records | length == 1
       - source_info.records.0.vcpu == cloned_info.records.0.vcpu
       - source_info.records.0.tags != cloned_info.records.0.tags
-      - source_info.records.0.boot_devices | length == cloned_info.records.0.boot_devices | length
       - source_info.records.0.disks | length ==  cloned_info.records.0.disks | length
       - source_info.records.0.nics | length ==  cloned_info.records.0.nics | length
       - source_info.records.0.nics.0.mac !=  cloned_info.records.0.nics.0.mac
       - source_info.records.0.node_affinity == cloned_info.records.0.node_affinity
+      # Cloned VM's boot devices should be set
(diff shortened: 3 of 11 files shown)


plugins/modules/vm_clone.py
```diff
@@ -193,7 +209,14 @@ def main():
...
                 options=dict(
-                    user_data=dict(type="str"),
+                    user_data=dict(
+                        type="str",
+                        default=(
+                            "runcmd:\n"
+                            "  - sed -i 's/^GRUB_DISABLE_LINUX_UUID=true/#GRUB_DISABLE_LINUX_UUID=true/' /etc/default/grub\n"
+                            "  - update-grub\n"
+                        ),
+                    ),
```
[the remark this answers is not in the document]
domendobnikar: Returned to the original module structure, added user_data "default" during payload creation.

plugins/module_utils/vm.py
```diff
@@ -610,6 +610,10 @@ def find_disk(self, slot):
             if disk.slot == slot:
                 return disk
 
+    # primary disk is the largest Virtio disk
+    def get_primary_disk(self):
+        return max(self.disk_list, key=lambda disk: disk.size)
```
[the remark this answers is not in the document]
domendobnikar: Added type

domendobnikar:
Integration tests: 
https://github.com/ScaleComputing/HyperCoreAnsibleCollection/actions/runs/27026276199/job/79767085521

plugins/module_utils/vm.py
```diff
@@ -423,6 +423,27 @@ def create_export_or_import_vm_payload(ansible_dict, cloud_init, is_export):
...
         return payload
 
+    @staticmethod
+    def clone_add_user_data_to_cloud_init(cloud_init):
+        # Task 370 - the generated cloud-init should include a runcmd
+        # to fix the grub UUID issue at first boot
+
+        user_data = cloud_init.get("user_data")
+        if not user_data:
+            return cloud_init
```
justinc1, line 433: No need to fix grub config in this case?

plugins/module_utils/vm.py
```diff
@@ -423,6 +423,27 @@ def create_export_or_import_vm_payload(ansible_dict, cloud_init, is_export):
...
+
+        cloud_init["user_data"] = (
+            user_data.rstrip()
+            + """
+
+        runcmd:
```
justinc1, line 439: We might end adding a user_data section when we already have a user_data section.
We can add a unit tests for those corner cases, then refaction this function a bit.
Corner cases:
- no user_data
- emtpy user_data
- no user_data.runcmd
- emtpy user_data.runcmd
- user_data.runcmd with entries

I guess we want to always comment out GRUB_DISABLE_LINUX_UUID ?

plugins/module_utils/vm.py
```diff
@@ -446,6 +467,7 @@ def create_clone_vm_payload(
                     hypercore_tags.append(tag)
             data["template"]["tags"] = ",".join(hypercore_tags)
         if cloud_init:
+            cloud_init = cls.clone_add_user_data_to_cloud_init(cloud_init)
```
justinc1, line 470: I have impression we want to alway run this?

tests/integration/targets/vm_clone/tasks/10_clone_stopped.yml
```diff
@@ -29,11 +29,12 @@
       - cloned_info.records | length == 1
       - source_info.records.0.vcpu == cloned_info.records.0.vcpu
       - source_info.records.0.tags != cloned_info.records.0.tags
-      - source_info.records.0.boot_devices | length == cloned_info.records.0.boot_devices | length
       - source_info.records.0.disks | length ==  cloned_info.records.0.disks | length
       - source_info.records.0.nics | length ==  cloned_info.records.0.nics | length
       - source_info.records.0.nics.0.mac !=  cloned_info.records.0.nics.0.mac
       - source_info.records.0.node_affinity == cloned_info.records.0.node_affinity
+      # Cloned VM's boot devices should be set
+      - cloned_info.records.0.boot_devices | length != 0
```
justinc1, line 37: I would prefer to test with "==".

I guess `source_info.records.0.boot_devices | length` == 0?

And `cloned_info.records.0.boot_devices | length` was 0 before, 1 after PR?

plugins/module_utils/vm.py
```diff
@@ -610,6 +632,13 @@ def find_disk(self, slot):
             if disk.slot == slot:
                 return disk
 
+    # primary disk is the largest Virtio disk
+    def get_primary_disk(self):
```
justinc1, line 636: We could clone a VM with IDE or SCSI disk(s).

Not sure why is the largest disk the bootable one. VM with 100 kB cloud-init ISO, 20 GB OS disk, 100 GB installed-apps disk. Maybe the first disk on bus is the OS disk, but if we add/remove disk, we likely can change that too. :( :(

Ideally we could ask HC3 about this setting for the source VM. But source VM might never be booted before clonning.

Or, we could clone VM, boot it, then assuming cloned VM does boot (really boot, not just wait on 
"no OS installed") set this from cloned VM runtime state.

A NIC could be a boot device too.

justinc1 (commented):
I think PR solves exactly the problem described in issue (e.g. virtio disk etc), but also creates additional corner cases.

I think we need to investigate what can be done. What if sorce VM has IDE or SCSI disk, a mixture, a NIC boot device etc.

I'm not sure we will be able to fully solve the problem. I tried to create a simple emtpy VM, with 1 disk, in HC3 web UI. The VM got beside the disk also a NIC, (empty) cloud-init CD-ROM, and scale-guest-tools-4.2.iso CD-ROM. The disk and  (empty) cloud-init CD-ROM were set as boot devices by HC3. In this case it seems like source VM boot order could be used as source of truth. But I did not check all potential ways to create a source VM.

plugins/module_utils/vm.py
domendobnikar: I'm not 100% sure, this is from the task description: "_If the module accepts cloud_init.user_data, the generated cloud-init should include a runcmd to fix the grub UUID issue at first boot_". 
I read it as "if the the user_data was passed"

plugins/module_utils/vm.py
domendobnikar: _a cloud-image VM cloned via this module boots correctly for years while it has exactly one virtio disk, then bricks on the first reboot after any tool (CSI driver, manual disk-add, Terraform provider, or even another playbook in the same suite) attaches a second virtio disk._ 
From my understanding this issue is a corner case to solve the "one virtio disk already exists and is used for boot but isn't specified as one and causes issues later on when another disk is added"

